PR #23, opened 2017-03-17
Fix sitemap locations property loading

ondrejsevcik:
Loading `ContentItemSystemAttributes` had a bug. Property `SitemapLocations` wasn't loaded because it's name wasn't found by Newtonsoft.Json (`sitemapLocations` in constructor vs. `sitemap_locations` in JSON).

I've fixed it by annotating properties with `JsonProperty` attribute. I've also added this attribute to all other models. It should remind us that property name and json name might not match.


petrsvihlik:
thx @ondrejsevcik . it's much cleaner this way

PR #38, opened 2017-04-25
Update DeliveryEndpointUrlBuilder.cs

makma:
Remove debug check


petrsvihlik:
@makma Could you provide more info on why did you remove it? Thx

makma:
This change was agreed and approved onDevOps meeting (by martinh and jirik). This change will help us (in the future) with testing of Compose on the staging environment.

petrsvihlik:
then we can merge it

Issue #35, opened 2017-04-13
Use IDeliveryClient wherever possible


petrsvihlik:
Places like `internal DeliveryItemResponse(JToken response, DeliveryClient client)` should use `IDeliveryClient`.

The concrete implementation - `DeliveryClient` - shouldn't be used in the SDK at all.

petrsvihlik:
fixed in the related PR

PR #42, opened 2017-05-10
fix: issue #35

jjarnot-vimanet:


jjarnot-vimanet:
First of all let me apologize for a late response.

For some reason DeliveryClient has got ContentLinkResolver property which is internal.
This is blocking me in using IDeliveryClient in the classes you mentioned above.

If you want I may change ContentLinkResolver property to be public and add it to IDeliveryClient.
After that I think I will be able to do what you want.
Are you ok with that?

petrsvihlik:
We shouldn't expose that member. I think it'll be just fine to move it to places where it's used (`ContentItem` and `CodeFirstModelProvider`) and instantiate it there with `IDeliveryClient.ContentLinkUrlResolver` as the only argument.

petrsvihlik:
fixed and merged

Issue #40, opened 2017-05-02
Assets don't include "Description" property


ChristopherJennings:
The asset type doesn't include the "description" property that is now available.

petrsvihlik:
implemented by @VarinderS

Issue #22, opened 2017-03-16
Replace usage of the sample project in tests with mocked responses


petrsvihlik:
Currently, the [`DeliveryClientTests`](https://github.com/Kentico/delivery-sdk-net/blob/master/KenticoCloud.Delivery.Tests/DeliveryClientTests.cs) depends on two sample projects in Kentico Cloud ("975bf280-fd91-488c-994c-2f04416e5ee3" and "e1167a11-75af-4a08-ad84-0582b463b010").

- [ ] We'd like to remove that dependency and use fake responses instead. Ideally, it should be done by mocking the httpclient (via moq or fakeiteasy).
- [ ] add test for #43 (`item.AssetField.First().Description`)

charithsoori:
Hi @petrsvihlik ,
I would like to pick this task. I hope this task will be a good starting point for me to delivery-sdk-net .

petrsvihlik:
Yes, indeed it's a good task to start with. Thanks for a great job! I've added a follow-up task https://github.com/Kentico/delivery-sdk-net/issues/84.

charithsoori:
Great . I'll start working on  #84

petrsvihlik:
fixed in #83

PR #43, opened 2017-05-12
Added description property to Asset model

VarinderS:
Implements @ChristopherJennings request #40

I'm wanting to add an assertion after [Line208](https://github.com/Kentico/delivery-sdk-net/blob/master/KenticoCloud.Delivery.Tests/DeliveryClientTests.cs#L208) in the ```DeliveryClientTests.cs``` file

But ```item.AssetField.First().Description``` property appears to be ```null``` for project with sandbox id ```e1167a11-75af-4a08-ad84-0582b463b010```

**Edit** looks like I messed up indentation, will update indentation to spaces


petrsvihlik:
thanks @VarinderS ! we'll fix the tests after #22 gets implemented

PR #53, opened 2017-06-20
Create test for preventing infinite loop during processing of circulary dependent inline content items

daemoon:


KenticoCloud.Delivery.Tests/CodeFirstModelProviderTests.cs
```diff
@@ -0,0 +1,108 @@
...
+            var retriever = new CodeFirstModelProvider(fakeDeliverClient);
+            retriever.TypeProvider = codeFirstTypeProvider;
+            A.CallTo(() => fakeDeliverClient.InlineContentItemsProcessor).Returns(processor);
+
+            var item = JToken.FromObject(rt1);
+            var modularContent = JToken.FromObject(modularContentObject);
+
+            var result = retriever.GetContentItemModel<ContentItemWithSingleRTE>(item, modularContent);
+
+            Assert.Equal("<span>FirstRT</span><span>SecondRT</span><span>FirstRT</span>", result.RT);
```
petrsvihlik: Is this really a correct behavior?
Shouldn't it be rather `<span>FirstRT</span><span>SecondRT</span>`?

KenticoCloud.Delivery.Tests/CodeFirstModelProviderTests.cs
```diff
@@ -0,0 +1,108 @@
...
+
+        private class RichTextInlineResolver: IInlineContentItemsResolver<ContentItemWithSingleRTE>
+        {
+            public string Resolve(ResolvedContentItemData<ContentItemWithSingleRTE> data)
+            {
+                return data.Item.RT;
+            }
+        }
+
+        private static object rt1 = new
```
petrsvihlik: Could you add an extra test for an items that directly reference themselves? RT3->RT3

petrsvihlik:
side note: internal reference = CMP-290

KenticoCloud.Delivery.Tests/CodeFirstModelProviderTests.cs
daemoon: No, the first one in which circular dependency is processed as if there were no inline content items. This is first detected at rt1 where we find that we're processing object which is already being processed.

KenticoCloud.Delivery.Tests/CodeFirstModelProviderTests.cs
daemoon: Yes

daemoon: Added

petrsvihlik:
thx

PR #60, opened 2017-07-11
Issue_#54 - add DeliveryOptions class to handle customer settings; mo…

jjarnot-vimanet:
Hi @petrsvihlik!

Please take a look at comments below:
1. Old constructors were removed from _DeliveryEndpointUrlBuilder_ - this internal class constructors were used only in _DeliveryClient_, so now they are not needed right?
2. _Configuration_ class  was removed since it is no longer used.
3. Instead of _IOptions<DeliveryOptions>_  we will be using _DeliveryOptions_ from now - it should not be a problem to extract _DeliveryOptions_ from _IOptions<>_ for the consumer
4. Logic from the following constructor 
_public DeliveryClient(string projectId)_
 was moved to the new one:
_public DeliveryClient(DeliveryOptions deliveryOptions)_

Thanks!


petrsvihlik:
1,2,4 - ok
3 - How will the configuration look from the consumer's point of view? Will it be possible to use `services.Configure<DeliveryOptions>(Configuration);` as outlined [here](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration)? we should follow the principle of least astonishment :)


What do you think @ChristopherJennings ?
Maybe one more thing - shouldn't we make `DeliveryOptions.ProjectId` `Guid` instead of `string`? It might be less consumer friendly though...

ChristopherJennings:
@petrsvihlik, I agree that the it should be able to work like you noted, and based on my (admittedly limited) understanding of that page and the proposed implementation, I _think_ it should work. I'd have to confirm. As for review, I won't likely be able to do any review this week, and I'm pretty solidly booked next week as well, but I can try for next week.

petrsvihlik:
Thanks @jjarnot-vimanet .
I made the following changes (https://github.com/Kentico/delivery-sdk-net/commit/a6645c98f38f9246710a858881ef8585d60236ad):
- added UsePreviewApi settings (to make it more clear for developers)
- refactored the constructors
- added extra constructor so that this will work: 

```
services.AddOptions();
services.Configure<DeliveryOptions>(Configuration);
services.AddSingleton<IDeliveryClient, DeliveryClient>();
```

PR #72, opened 2017-09-01
Feature/del 1192 add taxonomy endpoints

sutrkiller:
This should be published after the next release!

### DIFF
--- a/KenticoCloud.Delivery/DeliveryClient.cs
+++ b/KenticoCloud.Delivery/DeliveryClient.cs
@@ -417,6 +417,67 @@ namespace KenticoCloud.Delivery
             return new ContentElement(response, elementCodename);
         }
 
+
+        /// <inheritdoc />
+        public async Task<JObject> GetTaxonomyJsonAsync(string codename)
+        {
+            if (codename == null)
+            {
+                throw new ArgumentNullException(nameof(codename), "The codename of a taxonomy group is not specified.");
+            }
+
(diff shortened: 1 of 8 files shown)


KenticoCloud.Delivery/DeliveryClient.cs
```diff
@@ -417,6 +417,67 @@ public DeliveryClient(string projectId, string previewApiKey) : this(new Deliver
...
+            {
+                throw new ArgumentException("The codename of a taxonomy group is not specified.", nameof(codename));
+            }
+
+            var endpointUrl = UrlBuilder.GetTaxonomyUrl(codename);
+
+            return await GetDeliverResponseAsync(endpointUrl);
+        }
+
+        /// <inheritdoc />
```
petrsvihlik: I'm not a fan of duplicating code but in this case I'd vote for copying the XML docs from the interface. The reason for this is that VS doesn't support `<inheritdoc />`  in the intellisense.
Once it does we should switch to inherit doc and unify the style in the entire class.

petrsvihlik:
other than that, it looks fine

KenticoCloud.Delivery/DeliveryClient.cs
petrsvihlik: you can use r# to copy the comments. other than that, it looks fine

petrsvihlik:
ready to be merged once the endpoint is public (=test start passing)

PR #82, opened 2017-09-27
Copyright unicode sign

zenbhang:
added the unicode copyright sign


petrsvihlik:
sorry, we won't accept this pull request as there is no added value in it. moreover, the LICENSE file would cause a deviation from all other LICENSE files in our repositories.

PR #86, opened 2017-10-15
Added type filters

daemoon:
This is about the gist of it. Type will be generated based on codename field taken by reflection. Let me know if this is correct approach, and if anything else needs to be done.

### DIFF
--- a/KenticoCloud.Delivery/DeliveryClient.cs
+++ b/KenticoCloud.Delivery/DeliveryClient.cs
@@ -4,6 +4,7 @@ using System.Collections.Generic;
 using System.Net;
 using System.Net.Http;
 using System.Net.Http.Headers;
+using System.Reflection;
 using System.Threading.Tasks;
 using KenticoCloud.Delivery.InlineContentItems;
 using Microsoft.Extensions.Options;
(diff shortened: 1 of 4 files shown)


KenticoCloud.Delivery/DeliveryClient.cs
```diff
@@ -521,5 +525,24 @@ public DeliveryClient(string projectId, string previewApiKey) : this(new Deliver
 
             throw new DeliveryException(response.StatusCode, await response.Content.ReadAsStringAsync());
         }
+
+        private bool TryGetContentTypeCodename(Type contentType, out string codename)
```
petrsvihlik: I'd definitely put the whole thing in a separate class and unit-test it.
Also, consider making it an extension method. `parameters.ApplyTypeParameters<T>()`. But that's just a matter of taste.

KenticoCloud.Delivery/DeliveryClient.cs
```diff
@@ -247,7 +248,9 @@ public DeliveryClient(string projectId, string previewApiKey) : this(new Deliver
                 throw new ArgumentException("Entered item codename is not valid.", nameof(codename));
             }
 
-            var endpointUrl = UrlBuilder.GetItemUrl(codename, parameters);
+            var enhancedParams = ExtractParameters<T>(parameters);
```
petrsvihlik: it's redundant here - codename is more specific than type

daemoon:
Fixed those issues, please review

KenticoCloud.Delivery/DeliveryClient.cs
daemoon: FIxed

KenticoCloud.Delivery/DeliveryClient.cs
daemoon: Extracted into separate class which was unit tested. We can do it as extension method later.

PR #89, opened 2017-10-24
DEL-1250 handle long query string

petrpyszko:
see https://kentico.atlassian.net/browse/DEL-1250

### DIFF
--- a/KenticoCloud.Delivery/DeliveryClient.cs
+++ b/KenticoCloud.Delivery/DeliveryClient.cs
@@ -500,7 +500,7 @@ namespace KenticoCloud.Delivery
 
         private async Task<JObject> GetDeliverResponseAsync(string endpointUrl)
         {
-            var message = new HttpRequestMessage(HttpMethod.Get, endpointUrl);
+            var message = new HttpRequestMessage(HttpMethod.Get, GetUri(endpointUrl));
             if (_deliveryOptions.WaitForLoadingNewContent)
             {
                 message.Headers.Add("X-KC-Wait-For-Loading-New-Content", "true");
@@ -520,6 +520,18 @@ namespace KenticoCloud.Delivery
             }
 
             throw new DeliveryException(response.StatusCode, await response.Content.ReadAsStringAsync());
+        }
+
+        private Uri GetUri(string url)
+        {
+            try
+            {
+                return new Uri(url);
+            }
+            catch (UriFormatException ex)
+            {
+                throw new Exception("The request url is too long. Split your query into multiple calls.", ex);
+            }
         }
     }
 }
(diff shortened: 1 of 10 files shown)


KenticoCloud.Delivery/DeliveryClient.cs
```diff
@@ -521,5 +521,17 @@ public DeliveryClient(string projectId, string previewApiKey) : this(new Deliver
...
+
+        private Uri GetUri(string url)
+        {
+            try
+            {
+                return new Uri(url);
+            }
+            catch (UriFormatException ex)
+            {
+                throw new Exception("The request url is too long. Split your query into multiple calls.", ex);
```
petrsvihlik: I'm kinda missing the point here. Why to wrap the original exception? It already contains the information about the URL being too long. Or is it just to add the extra hint about splitting the query into multiple calls?

KenticoCloud.Delivery/DeliveryException.cs
```diff
@@ -23,11 +23,19 @@ public sealed class DeliveryException : Exception
...
             StatusCode = statusCode;
-            Message = JObject.Parse(message)["message"].ToString();
+
+            try
+            {
+                Message = JObject.Parse(responseStr)["message"].ToString();
+            }
+            catch (Exception)
+            {
+                Message = $"Unknown error. HTTP status code: {statusCode}.";
```
petrsvihlik: maybe add the `ReasonPhrase` from the response object? just to give the dev a little more insight...

KenticoCloud.Delivery/DeliveryClient.cs
petrpyszko: Yeah, I wanted to write some suggestion because the user of the SDK should know how to fix this problem.

KenticoCloud.Delivery/DeliveryException.cs
petrpyszko: Good point, I will add it.

KenticoCloud.Delivery/DeliveryClient.cs
petrsvihlik: right...on the other hand - you don't know if the UriFormatException will be caused by the length of the query. so your message could be completely off in some cases

Enngage: I would also not throw general `Exception`, but rather use something more specific

petrsvihlik:
it's fine by me. i think we can merge the changes...any objections @huluvu21 ?

petrpyszko:
- I've copypasted the hard limit from https://msdn.microsoft.com/en-us/library/z6c2z492(v=vs.110).aspx and we are throwing UriFormatException with more descriptive message just for long URLs

PR #90, opened 2017-10-25
DEL-1354 - Support of null values in "modular_content" within ritch text.

winklertomas:
Null values need to be supported in "modular_content" within ritch text.


petrsvihlik:
@winklertomas  already merged the changes. there's nothing to review in this PR

PR #96, opened 2018-02-01
Support for secured production Delivery API

### LINKED ISSUE #91: Add support for secured access to published (production) content
In coming weeks, the Delivery API will start supporting authenticated access.

In the first version of this feature, content will be secured using an API key. The developer will be able to turn the authenticated access on and off for the whole project. There will be one authentication key for the project, not multiple ones (for various user roles, for example).

Implement support for submitting the key in the form of the "Authorization" request header. The key should be sent via the [Bearer](https://swagger.io/docs/specification/authentication/bearer-authentication/) scheme (with a `"Bearer "` prefix).

We'll provide more information once they become available. https://kenticocloud.com/roadmap#secured-delivery-api

JanLenoch:
Fixes #91.

### DIFF
--- a/KenticoCloud.Delivery.Tests/DeliveryClientTests.cs
+++ b/KenticoCloud.Delivery.Tests/DeliveryClientTests.cs
@@ -1,14 +1,11 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.IO;
-using System.Net;
-using System.Net.Http;
-using System.Threading;
 using FakeItEasy;
 using Xunit;
 using RichardSzalay.MockHttp;
-using Moq;
+using System.IO;
+using System.Net;
 
 namespace KenticoCloud.Delivery.Tests
 {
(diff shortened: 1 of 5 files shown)


petrsvihlik:
@ChristopherJennings just fyi - it's been taken care of...
and the other SDKs:
- https://github.com/Kentico/delivery-sdk-java/issues/43
- https://github.com/Kentico/delivery-sdk-php/issues/32 
- https://github.com/Enngage/KenticoCloudDeliveryTypeScriptSDK/issues/42

Simply007:
I have rewritten, how does the SDK is sending request with virtual overload of the method.
Then I have used this virtual method to place the callback with assert before the actual request is send to Assert correct Authoritazion header value.

Same could be used for preview.

I haven't used cancellation token while sending (around 532 line in DeliveryClient.cs).
Take a look on it, it there is any best practice how to work with cancellation token - i have no experience with cancellation approach.

Simply007:
Test ```PreviewAndSecuredProductionThrows```
I think it is not necessary to test all the combination, if we want - we should consider to rename the test.

But this is a marginal note :-)

Simply007:
ad. 3. I have renamed test method.

ad 1. and 2. - created an issue https://github.com/Kentico/delivery-sdk-net/issues/97
Let's use either Moq or FakeItEasy...it should be possible to achieve what you're doing with FIE - https://fakeiteasy.readthedocs.io/en/stable/calling-base-methods/
Do not call live endpoint in tests

PR #98, opened 2018-02-05
Test correct value of the security header

### LINKED ISSUE #97: Try to use virtual method of httpclient for sdk for better maintainability
To be able to test the ```HttpRequestMessage``` easilly, we can use ```Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken);``` instead of current ```public Task<HttpResponseMessage> SendAsync(HttpRequestMessage request);```

We could use this pattern to test the message properties (i.e. headers).

I have found this possibility in:
https://github.com/Kentico/delivery-sdk-net/pull/96

Sample test that needs some changes is here:
https://github.com/Kentico/delivery-sdk-net/pull/96/commits/5a77c12cc3500aeba0e37c2f2cc2082fa189e18f

Fixes, that needs to be done:
1. Let's use either Moq or FakeItEasy...it should be possible to achieve what you're doing with FIE - https://fakeiteasy.readthedocs.io/en/stable/calling-base-methods/
2. Do not call live endpoint in tests

Simply007:
Fixes #97

### DIFF
--- a/KenticoCloud.Delivery.Tests/DeliveryClientTests.cs
+++ b/KenticoCloud.Delivery.Tests/DeliveryClientTests.cs
@@ -27,11 +27,11 @@ namespace KenticoCloud.Delivery.Tests
         {
             string url = $"{baseUrl}/items/";
 
-            mockHttp.When($"{url}{"coffee_beverages_explained"}").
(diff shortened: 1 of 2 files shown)


Simply007:
I have used already installed packages (FakeItEassy and RichardSzalay.MockHttp) and its Expect with headers functionality:
- https://github.com/richardszalay/mockhttp#matchers-with

PR #102, opened 2018-02-15
Fixes #99 - Add type filter only if missing

### LINKED ISSUE #99: Add type filter only if missing
https://github.com/Kentico/delivery-sdk-net/blob/b00094b2a5f633c4126e674607ecc3a7b3995966/KenticoCloud.Delivery/QueryParameters/Utilities/ContentTypeExtractor.cs#L15

JanLenoch:
Tested via the ExtractParameters_WhenGivenTypeWithCodenameAndExistingTypeParameter_DoesNotAddCodenameToParams test method.


Simply007:
I have extracted condition to the private method.
Also change condition operands order in order to load the codename only when it is required to store (not already in extended parameters)

PR #111, opened 2018-05-18
Update Readme.md

sedojj:
Strongly-typed retrieving examples missed generics, which is little confusing when trying to work with the example.


petrsvihlik:
Thx @sedojj

Issue #110, opened 2018-05-15
Add netstandard2.0 as a target


petrsvihlik:
**Problem:**
When installing the SDK, user can sometimes get this:
![microsoftteams-image](https://user-images.githubusercontent.com/9810625/40051002-826bab4a-5839-11e8-9804-0d6aa2a11863.png)
There are too many package dependencies. We try to reduce the number of references to the necessary minimum.

The root cause is actually targeting netstandard1.3: https://stackoverflow.com/questions/47365136/why-does-my-net-standard-nuget-package-trigger-so-many-dependencies. In [CM Helpers SDK](https://www.nuget.org/packages/KenticoCloud.ContentManagement.Helpers/) we're using netstandard2.0 + net45 targets and there's no such issue.

**Here's a recommendation from David Fowler:**
https://twitter.com/davidfowl/status/1008453902058917888

> I know we said target The lowest netstandard but we really want everyone to re-baseline at  netstandard 2.0. At the very least add a netstandard 2.0 target.

Other services use the same [approach](https://github.com/keenlabs/keen-sdk-net/blob/master/Keen/Keen.csproj) as well.

**Solution:**
- add netstandard2.0 target to reduce the dependency footprint of the SDK.

Additional steps we can take to reduce the dependency footprint are described in #117.

PeterSkoda:
I believe the root cause is actually because of targetting netstandard1.3, not because of net45 moniker. It is also mentioned here as well: https://stackoverflow.com/questions/47365136/why-does-my-net-standard-nuget-package-trigger-so-many-dependencies. In [CM Helpers SDK](https://www.nuget.org/packages/KenticoCloud.ContentManagement.Helpers/) we're using netstandard2.0 + net45 targets and there's no such issue (long list of dependencies) -

petrsvihlik:
Here's a recommendation from David Fowler:
https://twitter.com/davidfowl/status/1008453902058917888

> I know we said target The lowest netstandard but we really want everyone to re-baseline at  netstandard 2.0. At the very least add a netstandard 2.0 target.

Let's add netstandard2.0 target then.

PR #115, opened 2018-06-22
Fixes #100 - retry policy

### LINKED ISSUE #100: Implement a retry policy
Add a capability of dealing with transient network outages via the following retry policy:

* The retry policy should be **enabled** by default.
* The following should be configurable (preferably via the `DeliveryOptions` class):
  * enable/disable the retry logic (bool)
  * maximum amount of retry attempts (int)
* If specified, the Delivery client should retry in exponential retry intervals.
* The interval, in milliseconds, is always calculated as a power of the number 2. The exponent is the ordinal number of the next attempt. The result is then multiplied by 100. The order is one-based. For instance, the first retry attempt interval equals to (2<sup>2</sup> * 100), the second equals to (2<sup>3</sup> * 100), etc.
* When all retry attempts fail, the SDK should notify the client code about it, preferable via an exception.

The preferred way of implementing is via the https://github.com/App-vNext/Polly library (part of the .NET Foundation).

JanLenoch:
A Polly resilience policy according to specs.

The policy can be disabled via DeliveryOptions, also the maximum attempts can be configured. The implementation can be replaced with a custom-tailored policy plugged via DI into the constructor of the DeliveryClient. No breaking changes were introduced. The docu is updated.


Simply007:
My changes in general
* I have added a couple of new tests mainly for setting combinations
  * I have tries to make a tests that would check what return httpo status codes are handled b default by the retry policy, but the API does not ofer to get these status codes from the Policy object and run the fake reuests take 6s per one test, so i have decided to omit that test
* changed status codes array to the field
* ann note about property `ResiliencePolicyProvider` in readme

Simply007:
Just one note to the tests - I have descicded to not introduce some private helper methods - it makes the test not easy to read when you need to jump betwee the private methods and the tests which is calling them.

KenticoCloud.Delivery/DeliveryClient.cs
```diff
@@ -528,16 +560,27 @@ public DeliveryClient(string projectId, string previewApiKey) : this(new Deliver
...
-                var content = await response.Content.ReadAsStringAsync();
+                var content = await httpResponseMessage.Content?.ReadAsStringAsync();
 
                 return JObject.Parse(content);
             }
 
-            throw new DeliveryException(response, await response.Content.ReadAsStringAsync());
+            string faultContent = null;
+
+            // The null-coallescing operator cannot be used, hence "if" statement.
```
petrsvihlik: why is that? it seems to work for me....
either explain the reasons clearly in the comment or drop the comment

KenticoCloud.Delivery/KenticoCloud.Delivery.csproj
```diff
@@ -35,6 +35,7 @@
     <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="1.0.0" />
     <PackageReference Include="Microsoft.Extensions.Options" Version="1.0.0" />
     <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
+    <PackageReference Include="Polly" Version="6.0.1" />
```
petrsvihlik: I'm getting the following errors when opening the project.

Could you please find the reason why and try to fix it? Maybe it'd help to just switch to netstandard2.0 or updating some references. I'd be happy with either.


image

README.md
```diff
@@ -233,6 +233,19 @@ string transformedAssetUrl = builder.WithFocalPointCrop(560, 515, 2)
 
 For list of supported transformations and more information visit the Kentico Delivery API reference at <https://developer.kenticocloud.com/v1/reference?#image-transformation>.
 
+## Resilience capabilities
```
petrsvihlik: this whole chapter could use some copyediting. could you please create a task in the CTC backlog?

KenticoCloud.Delivery/DeliveryClient.cs
petrsvihlik: oh, the tests are failing...but still, please explain the reasons for the future generations

KenticoCloud.Delivery/DeliveryClient.cs
```diff
@@ -511,6 +522,27 @@ public DeliveryClient(string projectId, string previewApiKey) : this(new Deliver
...
+
+        private async Task<JObject> GetRepeatableDeliverResponseAsync(string endpointUrl)
+        {
+            var policyResult = await ResiliencePolicyProvider?.Policy?.ExecuteAndCaptureAsync(() =>
+                {
+                    return SendHttpMessage(endpointUrl);
+                }
+            );
+
+            return await GetResponseContent(policyResult?.FinalHandledResult ?? policyResult?.Result);
```
petrsvihlik: wouldn't it make sense to call `GetResponseContent` and its other occurrence in the `private async Task<JObject> GetDeliverResponseAsync(string endpointUrl)` method? (just askin')

KenticoCloud.Delivery/ResiliencePolicy/IResiliencePolicyProvider.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System.Net.Http;
+using Polly;
+
+namespace KenticoCloud.Delivery.ResiliencePolicy
+{
+    /// <summary>
+    /// Provides a resilience policy.
+    /// </summary>
+    public interface IResiliencePolicyProvider
```
petrsvihlik: i guess this is to make it more future-proof in case we want to add more policies, etc., right? I mean we could have injected `IAsyncPolicy<HttpResponseMessage>` directly in the `DeliveryClient` but this makes it more flexible. is that the idea?

KenticoCloud.Delivery/DeliveryClient.cs
```diff
@@ -511,6 +522,27 @@ public DeliveryClient(string projectId, string previewApiKey) : this(new Deliver
...
+        }
+
+        private async Task<JObject> GetSolitaryDeliveryResponseAsync(string endpointUrl)
```
petrsvihlik: solitary...why not just single? or omit the adjective completely? try to use understandable language, simple english.

KenticoCloud.Delivery/DeliveryClient.cs
JanLenoch: Done

KenticoCloud.Delivery/KenticoCloud.Delivery.csproj
JanLenoch: Sure, will do.

README.md
JanLenoch: CTC-256

KenticoCloud.Delivery/DeliveryClient.cs
JanLenoch: Yes, it would. I see this as a matter of personal preference. :)

KenticoCloud.Delivery/ResiliencePolicy/IResiliencePolicyProvider.cs
JanLenoch: Not really. The thing is that not only the `IAsyncPolicy<HttpResponseMessage>` object is required to use Polly. There's always some other wiring involved. Hence that wrapping provider.

KenticoCloud.Delivery/DeliveryClient.cs
JanLenoch: Done.

KenticoCloud.Delivery/DeliveryClient.cs
JanLenoch: Allright, alright. :) I've ditched these two methods completely (see above).

JanLenoch:
The comments from the review were incorporated via a new [branch](https://github.com/Kentico/delivery-sdk-net/tree/100-unresolvable-references). The code runs correctly in local environments but not in AppVeyor. I'd need to consult it with someone versed with AV.

petrsvihlik:
@JanLenoch I think some of the intended changes got into the #116 instead of this one

JanLenoch:
Once #116 is reviewed, could you @petrsvihlik or @Simply007 review this one? Thanks!

petrsvihlik:
Fixes #110 
Fixes #100

PR #120, opened 2018-08-15
Feature/KA-388 add tracking header for sdk

kubmir:
In this pull request, implementation of the requested tracking header, described in the [issue  #93](https://github.com/Kentico/delivery-sdk-net/issues/93), is added. As a part of the solution, the automatic generation of assembly information was added.

Internal ref: KA-388

### DIFF
--- a/KenticoCloud.Delivery/DeliveryClient.cs
+++ b/KenticoCloud.Delivery/DeliveryClient.cs
@@ -3,12 +3,12 @@ using System.Collections.Generic;
 using System.Net;
 using System.Net.Http;
 using System.Threading.Tasks;
+using Microsoft.Extensions.Options;
+using Newtonsoft.Json.Linq;
 
 using KenticoCloud.Delivery.Extensions;
 using KenticoCloud.Delivery.InlineContentItems;
 using KenticoCloud.Delivery.ResiliencePolicy;
-using Microsoft.Extensions.Options;
-using Newtonsoft.Json.Linq;
 
 namespace KenticoCloud.Delivery
 {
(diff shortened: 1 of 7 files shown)


KenticoCloud.Delivery/DeliveryClient.cs
```diff
@@ -272,10 +256,8 @@ public async Task<DeliveryItemResponse<T>> GetItemAsync<T>(string codename, IEnu
         /// </summary>
         /// <param name="parameters">An array that contains zero or more query parameters, for example, for filtering, ordering, or setting the depth of modular content.</param>
         /// <returns>The <see cref="DeliveryItemListingResponse"/> instance that contains the content items. If no query parameters are specified, all content items are returned.</returns>
-        public async Task<DeliveryItemListingResponse> GetItemsAsync(params IQueryParameter[] parameters)
-        {
-            return await GetItemsAsync((IEnumerable<IQueryParameter>)parameters);
-        }
+        public async Task<DeliveryItemListingResponse> GetItemsAsync(params IQueryParameter[] parameters) => 
+            await GetItemsAsync((IEnumerable<IQueryParameter>)parameters);
```
PetrSvirak: Would you mind moving the "arrow" on the line with expression? I know it is a tiny detail and a matter of habit but since we write symbols of the ternary operator and dots in longer expressions at the beginnings of lines, it seems a bit more natural to write the arrow on the new line as well to me.

kubmir:
Requested changes are implemented. 

I also added code-styles for both R# and visual studio - I will consult it with @PetrSvirak because I am not sure if it is a good idea. Maybe just visual studio .editorconfig is enough due to the overriding of R# rules.

KenticoCloud.Delivery/Extensions/HttpRequestHeadersExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+
+namespace KenticoCloud.Delivery.Extensions
+{
+    internal static class HttpRequestHeadersExtensions
+    {
+        private const string SdkTrackingHeaderName = "X-KC-SDKID";
+        private const string WaitForLoadingNewContentHeaderName = "X-KC-Wait-For-Loading-New-Content";
+
+        private const string PackageRepositoryHost = "nuget.org";
+        private const string SdkPackageId = "KenticoCloud.Delivery";
```
petrsvihlik: i think this can be safely extracted from assembly information as well (assembly.GetName().Name)

petrsvihlik:
Fixes #93

PR #125, opened 2018-09-10
fixes #124

### LINKED ISSUE #124: Circular Dependency is resulting in a overflow exception
### Brief bug description

I have a content model with a richtext component.  Based on this content model an user creates a content item in the KC portal. In the richtext component the user is able to refer to other content items. 
If the user makes the content item referencing itself (Circular Dependency) the KenticoCloud.Delivery  SDK is unable to deserialize the content item and throws an overflow exception .

### Repro steps
See above, Example repo is avilable @ https://github.com/micheltol/Kentico_CircularDependency

### Expected behavior
I would expect this would not be possible in the KC portal. 
Apperently there is an usecase for this behavior so I would expect the SDK to be able to deserialize the above mentioned situation.

### Test environment

 - Platform/OS: .NET Core 2.1
 - Version  Nuget KenticoCloud.Delivery : 5.0.0

petrsvihlik:
### Motivation

Which issue does this fix? Fixes #124

### Checklist

- [x] Tests are passing
- [x] Docu has been updated (if applicable)

### How to test

This repo: https://github.com/micheltol/Kentico_CircularDependency should work after the client is updated to the latest version.


Simply007:
I have split json files that are used as a source when using mock http client.
This change prevents to fail both tests when only one of the functionality breaks.
+
Also unified all json source files to be copied always when the build is performed.

Simply007:
Fix is OK for me.

PR #131, opened 2018-10-01
Added support for content components

petrsvihlik:
### Motivation

The goal is to add support for content components (non-reusable content). 
From the SDKs point of view, the components behave the same as regular content items. However, they are supported only in the rich-text elements. 

`<object type="application/kenticocloud" data-type="component" data-codename="n55afed3a_b4a3_018a_8744_3b06301b1f88"></object>`

The only difference is that the value of the `data-type` attribute is not `item` but `component`. (There is no new endpoint for Content components in the Delivery API). The components are being sent as a part of the modular content (linked items) dictionary in the same format as the regular modular content. 

### Checklist

- [x] Tests are passing
- [ ] Temporary settings (e.g. project ID used during development and testing) have been reverted to defaults

### How to test

Make sure rich-text elements containing components get resolved in the same manner as standard content items.
Testing KC project: https://qa-deliver.global.ssl.fastly.net/ab44e0cc-9302-00c3-14ce-1362a118ad2f/items

_Internal reference: DC-544_


petrsvihlik:
solved in a different way. support from the SDK side is not needed.

Issue #135, opened 2018-10-12
Add an explicit exception message about network outage


JanLenoch:
### Motivation

The most common reason of the execution flowing through the whole body of DeliveryClient.GetResponseContent() is that either the retry policy was disabled or all retry attempts were depleted.

### Proposed solution

Add an exception message with such a hint.

petrsvihlik:
fixed in #137

PR #137, opened 2018-10-18
Added a hint about network outage to the exception message at the end…

RichardKalinec:
The more explicit exception message will be more helpful for understanding possible reasons of network outage. Before printing the fault content, the message now prints out: "Either the retry policy was disabled or all retry attempts were depleted."


petrsvihlik:
fixes #135

PR #147, opened 2018-11-28
fixes #146

### LINKED ISSUE #146: Resolving inline items with IInlineContentItemsResolver is broken in v8
### Brief bug description

After upgrading SDK to v8, all non-default `IInlineContentItemsResolver` registered using the `DeliveryClientBuilder` with `WithInlineContentItemsResolver` such as:

```
    public class InlineCodeResolver : IInlineContentItemsResolver<CodeBlock>
    {
        public string Resolve(ResolvedContentItemData<CodeBlock> data)
        {
            return
                $"Do something";
        }
    }

```
` .WithInlineContentItemsResolver(new InlineCodeResolver())`

Are not resolved and never called. These resolvers worked in previous versions as expected.

The only resolver that gets called is the default one registered with `object`. Strangely, default resolver gets strongly typed items as I expect so I can do something like this as a workaround:

```
    {
        {
            if (data.Item is CodeBlock codeBlock)
            {
                return
                    $"Do something";
            }
        }
    }
```

### Expected behavior

Registered resolvers should be called instead of falling back to default one.

petrsvihlik:
### Motivation

Which issue does this fix? Fixes #146

### Checklist

- [x] Tests are passing

### How to test

Follow the repro steps in #146 

Note: there's a breaking change - I removed some methods from one interface. The methods didn't really make sense there.

### DIFF
--- a/KenticoCloud.Delivery/InlineContentItems/IInlineContentItemsProcessor.cs
+++ b/KenticoCloud.Delivery/InlineContentItems/IInlineContentItemsProcessor.cs
@@ -6,11 +6,7 @@ namespace KenticoCloud.Delivery.InlineContentItems
     /// Interface implemented for processing inline content items in HTML
     /// </summary>
     public interface IInlineContentItemsProcessor
-{        /// <summary>
-        /// Gets or sets resolver used in case no other resolver was registered for type of inline content item
-        /// </summary>
-        IInlineContentItemsResolver<object> DefaultResolver { get; set; }
-
+    {
         /// <summary>
         /// Processes HTML input and returns it with inline content items replaced with resolvers output.
         /// </summary>
@@ -18,14 +14,7 @@ namespace KenticoCloud.Delivery.InlineContentItems
         /// <param name="usedContentItems">Content items referenced as inline content items</param>
         /// <returns>HTML with inline content items replaced with resolvers output</returns>
         string Process(string value, Dictionary<string, object> usedContentItems);
-
-        /// <summary>
-        /// Function used for registering content type specific resolvers used during processing.
-        /// </summary>
-        /// <param name="resolver">Method which is used for specific content type as resolver.</param>
-        /// <typeparam name="T">Content type which is resolver resolving.</typeparam>
-        void RegisterTypeResolver<T>(IInlineContentItemsResolver<T> resolver);
-
+        
         /// <summary>
         /// Removes all content items from given HTML content.
         /// </summary>
--- a/KenticoCloud.Delivery/InlineContentItems/IInlineContentItemsResolverCollection.cs
+++ b/KenticoCloud.Delivery/InlineContentItems/IInlineContentItemsResolverCollection.cs
@@ -1,0 +1,16 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace KenticoCloud.Delivery.InlineContentItems
+{
+    /// <summary>
+    /// Collection of inline content item resolvers indexed by types that they resolve.
+    /// </summary>
+    internal interface IInlineContentItemsResolverCollection
+    {
+        /// <summary>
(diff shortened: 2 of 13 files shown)


KenticoCloud.Delivery/InlineContentItems/IInlineContentItemsResolverCollection.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace KenticoCloud.Delivery.InlineContentItems
+{
+    /// <summary>
+    /// Collection of inline content item resolvers indexed by types that they resolve.
+    /// </summary>
+    internal interface IInlineContentItemsResolverCollection
```
PetrSvirak: That's the culprit!

KenticoCloud.Delivery/InlineContentItems/InlineContentItemsProcessor.cs
```diff
@@ -1,17 +1,30 @@
...
     /// <summary>
     /// Processor responsible for parsing HTML input and resolving inline content items referenced in them using registered resolvers
     /// </summary>
     internal class InlineContentItemsProcessor : IInlineContentItemsProcessor
     {
-        private readonly Dictionary<Type, Func<object, string>> _typeResolver;
+        private  Dictionary<Type, ResolveInlineContent> _inlineContentItemsResolvers;
+        private Dictionary<Type, ResolveInlineContent> InlineContentItemsResolvers
+        {
+            get => _inlineContentItemsResolvers ?? (_inlineContentItemsResolvers = new Dictionary<Type, ResolveInlineContent>());
```
PetrSvirak: That is a bit nasty. Since we use DI and we have the control over the default instance, we should probably not be over-protective towards dependencies…

As for tests, I believe, a factory that would ensure create instantiation in tests would be a bit better fitting 👼

KenticoCloud.Delivery/InlineContentItems/InlineContentItemsResolverCollection.cs
```diff
@@ -0,0 +1,22 @@
...
+namespace KenticoCloud.Delivery.InlineContentItems
+{
+    internal class InlineContentItemsResolverCollection : IInlineContentItemsResolverCollection
+    {
+        private Dictionary<Type, ResolveInlineContent> _inlineContentItemsResolvers;
+        public Dictionary<Type, ResolveInlineContent> InlineContentItemsResolvers
+        {
+            get
+            {
+                return _inlineContentItemsResolvers ?? (_inlineContentItemsResolvers = new Dictionary<Type, ResolveInlineContent>());
```
PetrSvirak: I would go [`Lazy`](https://docs.microsoft.com/en-us/dotnet/api/system.lazy-1?view=netframework-4.7.2) here…

KenticoCloud.Delivery/InlineContentItems/IInlineContentItemsProcessor.cs
```diff
@@ -6,26 +6,15 @@ namespace KenticoCloud.Delivery.InlineContentItems
...
         /// <param name="usedContentItems">Content items referenced as inline content items</param>
         /// <returns>HTML with inline content items replaced with resolvers output</returns>
         string Process(string value, Dictionary<string, object> usedContentItems);
-
-        /// <summary>
-        /// Function used for registering content type specific resolvers used during processing.
-        /// </summary>
-        /// <param name="resolver">Method which is used for specific content type as resolver.</param>
-        /// <typeparam name="T">Content type which is resolver resolving.</typeparam>
-        void RegisterTypeResolver<T>(IInlineContentItemsResolver<T> resolver);
```
PetrSvirak: 👍

PR #148, opened 2018-11-28
fixes #146

### LINKED ISSUE #146: Resolving inline items with IInlineContentItemsResolver is broken in v8

PetrSvirak:
### Motivation

Which issue does this fix? Fixes #146 

* Non-generic `IInlineContentItemsResolver` was removed as it was an unusable interface (breaking change) that might eventually cause confusion in combination with `ITypelessInlineContentItemsResolver`. 
* An independent `ITypelessInlineContentItemsResolver` interface was introduced to wrap and hide `IInlineContentItemsResolver<TContentItem>` resolvers and provide them in a non-generic way to the `InlineContentItemsProcessor` that later puts inline content items through the resolver in order to obtain their string representation – I've chosen this way to mitigate a breaking change in `IInlineContentItemsResolver<>` (aka all existing resolvers).
* New extension methods for IServiceCollection were introduced to allow registration of custom inline content items resolvers (docs needs to be eventually updated).
* I have added some tests for the `IInlineContentItemsResolver<>`s to the DI framework integration tests so we can be sure that the approach would work.

### Checklist

- [x] Tests are passing

### How to test

Follow the repro steps in #146

Note: there's another breaking change - Petr removed some methods from one interface. The methods didn't really make sense there (see https://github.com/Kentico/delivery-sdk-net/pull/147).

### DIFF
--- a/KenticoCloud.Delivery.Tests/Builders/DeliveryClient/DeliveryClientBuilderTests.cs
+++ b/KenticoCloud.Delivery.Tests/Builders/DeliveryClient/DeliveryClientBuilderTests.cs
@@ -54,8 +54,7 @@ namespace KenticoCloud.Delivery.Tests.Builders.DeliveryClient
             var mockInlineContentItemsProcessor = A.Fake<IInlineContentItemsProcessor>();
             var mockInlineContentItemsResolver = A.Fake<IInlineContentItemsResolver<object>>();
             var mockCodeFirstTypeProvider = A.Fake<ICodeFirstTypeProvider>();
-            var mockHttp = new MockHttpMessageHandler().ToHttpClient();
-            A.CallTo(() => mockInlineContentItemsProcessor.DefaultResolver).Returns(mockInlineContentItemsResolver);
+            var mockHttp = new MockHttpMessageHandler().ToHttpClient();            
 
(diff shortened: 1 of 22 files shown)


petrsvihlik:
awesomely done! my hope is that my PR at least gave you a head start :)

I was not aware of the black magic that's going on when you expect an IEnumerable of T in a constructor...makes things a lot simpler ;)

Issue #149, opened 2018-12-05
Custom default inline content items resolver registered through builder is not used


matus12:
### Brief bug description

When I register a custom default inline content items resolver and a resolver for UnretrievedContentItems through the DeliveryClientBuilder, the default implementation of the resolvers is used in the InlineContentItemsProcessor instead of the custom ones.

### Expected behavior

When I register a custom default inline content items resolver or the resolver for unretrieved content items DeliveryClientBuilder, they are used in the InlineContentItemsProcessor.

petrsvihlik:
will be released in https://github.com/Kentico/delivery-sdk-net/releases/tag/9.0.0

PR #150, opened 2018-12-05
Allow registering of default inline content items resolver

### LINKED ISSUE #145: Missing content model causes DeliveryClient to silently fail
### Brief bug description

When a new content model is created in KC and there is no corresponding POCO in the app (therefore CustomTypeProvider does not recognize the model), DeliveryClient fails when returning content items that contain content items based on this model. It does not throw any exception, but returns with Items property containing null.
Problem is in InlineContentItemsProcessor.cs on line 69 -> inlineContentItem is null, therefore invoking GetType() will generate NullReferenceException. This exception is not propagated to the app, but silently ignored and DeliveryClient just returns null in DeliveryItemListingResponse.

### Repro steps

1. Generate POCO for KC content models
2. Add new content model in KC
3. Create content item that has a rich text editor. Add linked item based on the new content model
4. Get the new content item using DeliveryClient.GetItemsAsync
5. Items property of the result is null

### Expected behavior

I'd expect either null or exception. Not null of the whole Items property, but of the linked item that there is no POCO for.

### Test environment

 - Platform/OS: .NET 4.6

matus12:
Includes fix of a bug where content items without model made DeliveryClient to fail.

### Motivation

Fixes #145 & #149 

### Checklist

- [x] Tests are passing


KenticoCloud.Delivery/InlineContentItems/InlineContentItemsProcessor.cs
```diff
@@ -121,5 +90,36 @@ private static List<IElement> GetInlineContentItemElements(AngleSharp.Dom.Html.I
...
+                return inlineContentItemResolver(inlineContentItem);
+            }
+
+            var data = new ResolvedContentItemData<object> { Item = inlineContentItem };
+            if (_inlineContentItemsResolvers.TryGetValue(typeof(object), out var defaultContentItemResolver))
+            {
+                return defaultContentItemResolver(data);
+            }
+
+            throw new NotSupportedException("Default inline content item resolver for non specific content type was not registered.");
```
petrsvihlik: I guess this addresses #145 .
What would you say if we did the same as in the `ReplaceElementWithFragmentNodes` method? I mean we could return the error within the text.
This would let the developer know something's wrong but the application wouldn't break completely when a new content type is added/used...
@matus12 @PetrSvirak @ondrabus ?

KenticoCloud.Delivery.Tests/ContentLinkResolverTests.cs
```diff
@@ -125,7 +125,10 @@ public async void ResolveLinksInStronglyTypedModel()
             var httpClient = mockHttp.ToHttpClient();
             var resiliencePolicyProvider = new DefaultResiliencePolicyProvider(deliveryOptions);
             var contentLinkUrlResolver = new CustomContentLinkUrlResolver();
-            var contentItemsProcessor = InlineContentItemsProcessorFactory.Create(new ReplaceWithEmptyStringResolver(), new ReplaceWithEmptyStringForUnretrievedItemsResolver());
+            var contentItemsProcessor = InlineContentItemsProcessorFactory
+                .WithResolver<object, ReplaceWithEmptyStringResolver>()
+                .AndResolver<UnretrievedContentItem, ReplaceWithEmptyStringForUnretrievedItemsResolver>()
```
petrsvihlik: Shouldn't we adjust the `ServiceCollectionExtensions` as well? And change 

https://github.com/Kentico/delivery-sdk-net/blob/master/KenticoCloud.Delivery/Extensions/ServiceCollectionExtensions.cs#L99
from
```
 services.TryAddSingleton<IInlineContentItemsResolver<object>, ReplaceWithWarningAboutRegistrationResolver>();
            services.TryAddSingleton<IInlineContentItemsResolver<UnretrievedContentItem>, ReplaceWithWarningAboutUnretrievedItemResolver>();
          
```

to 
`AddDeliveryInlineContentItemsResolver`
?

matus12: I believe you are right, I have tested it and the way we register it now doesn't work properly, using `AddDeliveryInlineContentItemsResolver` works.

petrsvihlik: cool, ideally, cover it with a test

matus12:
Ready for review

matus12:
Ready for review

matus12:
Ready for review

KenticoCloud.Delivery/CodeFirst/UnknownContentItem.cs
```diff
@@ -0,0 +1,21 @@
...
+    /// Type used to identify inline content items which don't have corresponding model.
+    /// </summary>
+    public class UnknownContentItem
+    {
+        /// <summary>
+        /// Represents the content type that has no corresponding model
+        /// </summary>
+        public string Type { get; set; }
+
+        /// <inheritdoc />
```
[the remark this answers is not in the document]
matus12: ReSharper said if I did that he would stop underlining the constructor 👼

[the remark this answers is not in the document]
PetrSvirak: [I see what you did there](https://imgflip.com/i/2orspv) 😆

KenticoCloud.Delivery.Tests/CodeFirstModelProviderTests.cs
```diff
@@ -35,6 +36,31 @@ public void RetrievingContentModelWithCircularDependencyDoesNotCycle()
...
+        public void RetrievingNonExistentContentModelCreatesWarningInRichtext()
+        {
+            var codeFirstTypeProvider = A.Fake<ICodeFirstTypeProvider>();
+            var contentLinkUrlResolver = A.Fake<IContentLinkUrlResolver>();
+            var propertyMapper = A.Fake<ICodeFirstPropertyMapper>();
+            A.CallTo(() => codeFirstTypeProvider.GetType(A<string>._)).Returns(null);
+            A.CallTo(() => propertyMapper.IsMatch(A<PropertyInfo>._, A<string>._, A<string>._)).Returns(true);
+
+            var processor = InlineContentItemsProcessorFactory
+                .WithResolver<UnknownContentItem, ReplaceWithWarningAboutNoModelResolver>()
```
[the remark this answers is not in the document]
matus12: Is dependency on `UnknownContentItem` alright? Or should there be a new class for that too?

KenticoCloud.Delivery.Tests/ContentLinkResolverTests.cs
```diff
@@ -125,7 +126,7 @@ public async void ResolveLinksInStronglyTypedModel()
             var httpClient = mockHttp.ToHttpClient();
             var resiliencePolicyProvider = new DefaultResiliencePolicyProvider(deliveryOptions);
             var contentLinkUrlResolver = new CustomContentLinkUrlResolver();
-            var contentItemsProcessor = InlineContentItemsProcessorFactory.Create(new ReplaceWithEmptyStringResolver(), new ReplaceWithEmptyStringForUnretrievedItemsResolver());
+            var contentItemsProcessor = InlineContentItemsProcessorFactory.Create();
```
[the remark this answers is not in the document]
matus12: I believe it is, they were there but had no impact on the test.

KenticoCloud.Delivery.Tests/CodeFirstModelProviderTests.cs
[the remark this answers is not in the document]
PetrSvirak: I believe so. All `object`, `UnknownContentItem`, `UnresolvedContentItem` are special types for us (either last resort or `provider`-generated under special conditions). While we internally treat all (but, to some extend, `object`'s) resolvers equally, we need to be sure that our default processor treats them well.

matus12:
Your changes make sense to me. @petrsvihlik ready for your review.

KenticoCloud.Delivery/CodeFirst/CodeFirstModelProvider.cs
```diff
@@ -308,6 +309,10 @@ private string ProcessInlineContentItems(JToken linkedItems, Dictionary<string,
                             contentItem = GetContentItemModel(typeof(object), linkedItemsElementNode, linkedItemsNode, processedItems, currentlyResolvedRichStrings);
                             if (!processedItems.ContainsKey(codenameUsed))
                             {
+                                if (contentItem == null)
+                                {
+                                    contentItem = new UnknownContentItem(linkedItemsElementNode);
```
petrsvihlik: Shouldn't we do this on the line 79 and react appropriately here?
```
if (t == null)
            {
                return null; // -> return new UnknownContentItem(item); ???
            }
```

KenticoCloud.Delivery/InlineContentItems/InlineContentItemsProcessor.cs
```diff
@@ -12,29 +13,29 @@ namespace KenticoCloud.Delivery.InlineContentItems
...
         private readonly HtmlParser _strictHtmlParser;
 
-        /// <summary>
-        /// Resolver used in case no other resolver was registered for type of inline content item
-        /// </summary>
-        public IInlineContentItemsResolver<object> DefaultResolver { get; set; }
-
         // Used by tests only
-        internal IEnumerable<Type> ContentItemTypesWithResolver => _inlineContentItemsResolvers.Keys;
+        internal IEnumerable<Type> ContentItemTypesWithResolver => _inlineContentItemsResolvers.Keys.ToArray();
```
petrsvihlik: I don't like "for tests only" :)
Why not to wrap the private field in an internal property and leave it that way without additional comments?

KenticoCloud.Delivery/CodeFirst/CodeFirstModelProvider.cs
matus12: The public GetContentItemModel method would try to change the type from UnknownContentItem to T and that would cause an exception.

KenticoCloud.Delivery/InlineContentItems/InlineContentItemsProcessor.cs
PetrSvirak: Well, it already is an internal property 😉. But it should not be used within the Delivery project itself as it was only meant for tests. If you think the comment is redundant and property won't be misused, let's remove the comment.

KenticoCloud.Delivery/CodeFirst/CodeFirstModelProvider.cs
petrsvihlik: hmm...good point.

KenticoCloud.Delivery/InlineContentItems/InlineContentItemsProcessor.cs
petrsvihlik: yeah, i can see that, thank you very much :D 
what I meant is - why not to convert it into an automatic property and do all the `.Keys` and `.ToArray()`s calls outside, in the tests to be precise... now it really feels like a single-purpose solution.

PetrSvirak: I still miss something here, I guess - it is an automatic property, isn't it? 😟 

The `.ToArray` should be here to prevent manipulation with the inner component state. But if you feel like making the `_inlineContentItemsResolvers` into an internal automatic property (is this what you meant [all along](https://memegenerator.net/img/instances/81112171.jpg)? 😊), then go for it. 

I would still prefer a backing field with an internal property that would make a [read-only copy](https://docs.microsoft.com/en-us/dotnet/api/system.collections.objectmodel.readonlydictionary-2?view=netframework-4.7.2) before providing the collection outside of the `Processor`. Something like:
```suggestion
        internal IReadOnlyDictionary<Type> ContentItemResolvers => new ReadOnlyDictionary<Type, Func<object, string>>(_inlineContentItemsResolvers);
```
But I will leave this completely upon you.

petrsvihlik: see, I was wonder why did you convert the collection to an array. I'll convert it to a readonly collection. this way the intentions will be much clearer :)

Neytus:
Fixes #149

PR #154, opened 2019-02-22
Inline content resolver in v9 has unnecessary wrapping (KA-520)

### LINKED ISSUE #153: Inline content resolver in v9 has unnecessary wrapping
This related to https://github.com/Kentico/delivery-sdk-net/issues/146 (which GitHub cant display for some reason). 

Basically, I have inline resolver defined as

```
    {

        {
            if (data.Item is ResolvedContentItemData<object> itemWrapper)
            {
                if (itemWrapper.Item is InfoBox infoBox)
                {
                    return $"<div>{infoBox.Content}</div>";
                }

                if (itemWrapper.Item is CodeBlock codeBlock)
                {
                    return
                        $"<pre><code class=\"language-{codeBlock.AvailableLanguagesLanguage?.FirstOrDefault()?.Codename.ToLower().Trim()}\">" +
                            $"{System.Web.HttpUtility.HtmlEncode(codeBlock.Code)?.Trim()}" +
                        $"</code></pre>";
                }
            }

        }
    }
```

Main problem I have is with the `Resolve(ResolvedContentItemData<object> data)` because I would expect the `Item` inside to be of resolved content item type (e.g. `CodeBlock` or `InfoBox` in my case), but its type is another `ResolvedContentItemData<object>` which is why I have to do:

```
if (data.Item is ResolvedContentItemData<object> itemWrapper)
    {
         if (itemWrapper.Item is InfoBox infoBox)
         {
         }
}
```

PetrSvirak:
### Motivation

Fixes #153

### Checklist

- [x] Tests are passing

### How to test

I suggest going through the scenario described in #153 and quickly check that [cloud-sample-app-net](https://github.com/Kentico/cloud-sample-app-net) would work as expected.

### DIFF
--- a/KenticoCloud.Delivery/InlineContentItems/InlineContentItemsProcessor.cs
+++ b/KenticoCloud.Delivery/InlineContentItems/InlineContentItemsProcessor.cs
@@ -102,19 +102,16 @@ namespace KenticoCloud.Delivery.InlineContentItems
                 return inlineContentItemResolver(inlineContentItem);
             }
 
-            var data = new ResolvedContentItemData<object> { Item = inlineContentItem };
             if (_inlineContentItemsResolvers.TryGetValue(typeof(object), out var defaultContentItemResolver))
             {
-                return defaultContentItemResolver(data);
+                return defaultContentItemResolver(inlineContentItem);
             }
 
             return "Default inline content item resolver for non specific content type was not registered.";
         }
 
         private static Type GetInlineContentItemType(object inlineContentItem)
-        {
-            return inlineContentItem?.GetType() ?? typeof(UnknownContentItem);
-        }
+            => inlineContentItem?.GetType() ?? typeof(UnknownContentItem);
 
         private void ReplaceElementWithFragmentNodes(AngleSharp.Dom.Html.IHtmlDocument document, IElement inlineContentItemElement, string contentItemCodename, object inlineContentItem, string fragmentText)
         {
--- a/KenticoCloud.Delivery.Tests/CodeFirstModelProviderTests.cs
+++ b/KenticoCloud.Delivery.Tests/CodeFirstModelProviderTests.cs
@@ -20,20 +20,20 @@ namespace KenticoCloud.Delivery.Tests
             var contentLinkUrlResolver = A.Fake<IContentLinkUrlResolver>();
             var propertyMapper = A.Fake<ICodeFirstPropertyMapper>();
             A.CallTo(() => propertyMapper.IsMatch(A<PropertyInfo>._, A<string>._, A<string>._)).Returns(true);
-            A.CallTo(() => codeFirstTypeProvider.GetType(A<string>._)).Returns(typeof(ContentItemWithSingleRTE));
+            A.CallTo(() => codeFirstTypeProvider.GetType(A<string>._)).Returns(typeof(ContentItemWithSingleRte));
 
             var processor = InlineContentItemsProcessorFactory
-                .WithResolver<ContentItemWithSingleRTE,RichTextInlineResolver>()
+                .WithResolver(ResolveItemWithSingleRte)
                 .Build();
             var retriever = new CodeFirstModelProvider(contentLinkUrlResolver, processor, codeFirstTypeProvider, propertyMapper);
 
-            var item = JToken.FromObject(rt1);
-            var linkedItems = JToken.FromObject(linkedItemsForItemWithTwoReferencedContentItems);
+            var item = JToken.FromObject(Rt1);
+            var linkedItems = JToken.FromObject(LinkedItemsForItemWithTwoReferencedContentItems);
 
-            var result = retriever.GetContentItemModel<ContentItemWithSingleRTE>(item, linkedItems);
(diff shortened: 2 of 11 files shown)


PetrSvirak:
I have done basically 3 things:
* fixed the issue
* refactored `InlineContentItem-` factories in tests to make touched tests a bit more readable
* internally reorganized/split `GetContentItemModel` method to wrap my head around the root cause of the issue. 

Neither of refactorings is completely necessary, and anybody finds them (well, only 334bbe0, really) too risky, I can remove the commit.

As for the issue, it technically is a breaking change, however, I believe we would release the version as a minor since this was the intended behavior in the first place.

If we agree on a new major, though, I would continue with refactoring a bit further and removed `ResolvedContentItemData` completely (definitely a breaking change, of course) and maybe even tried to figure out an even simpler way to create simple strongly typed inline resolvers (this would require a bit more validation – maybe [switch with pattern matching](https://docs.microsoft.com/en-us/dotnet/csharp/pattern-matching#using-pattern-matching-switch-statements) is our best bet anyway).

KenticoCloud.Delivery/InlineContentItems/InlineContentItemsProcessor.cs
```diff
@@ -102,19 +102,16 @@ private string GetFragmentText(object inlineContentItem)
                 return inlineContentItemResolver(inlineContentItem);
             }
 
-            var data = new ResolvedContentItemData<object> { Item = inlineContentItem };
             if (_inlineContentItemsResolvers.TryGetValue(typeof(object), out var defaultContentItemResolver))
             {
-                return defaultContentItemResolver(data);
```
petrsvihlik: I consider this a breaking change even though we're not changing the contract...

Enngage:
> I have done basically 3 things:
> 
> * fixed the issue
> * refactored `InlineContentItem-` factories in tests to make touched tests a bit more readable
> * internally reorganized/split `GetContentItemModel` method to wrap my head around the root cause of the issue.
> 
> Neither of refactorings is completely necessary, and anybody finds them (well, only [334bbe0](https://github.com/Kentico/delivery-sdk-net/commit/334bbe077c1a8efe6f0781724717b99204856282), really) too risky, I can remove a commit or two.
> 
> As for the issue, it technically is a breaking change, however, I believe we would release the version as a minor since this was the intended behavior in the first place.
> 
> If we agree on a new major, though, I would continue with refactoring a bit further and removed `ResolvedContentItemData` completely (definitely a breaking change, of course) and maybe even tried to figure out an even simpler way to create simple strongly typed inline resolvers (this would require a bit more validation – maybe [switch with pattern matching](https://docs.microsoft.com/en-us/dotnet/csharp/pattern-matching#using-pattern-matching-switch-statements) is our best bet anyway).

@PetrSvirak , I would personally hate to see the use of switch pattern here. If you had a switch, you would still had to cast the item to your proper type in order to work with it anyway. I feel that using `is` is superior and would leave it at that.

petrsvihlik:
Yes, we'll do a new major version. Could you please:
- split the refactorings you mentioned into smaller issues
- add them to the vNext milestone right after you create them (I'll rename it to 10.0.0)
- work on one issue at a time
  - once you pick up an issue, don't forget to assign it to yourself

I suggest we do this as well (while we have the chance):
- https://github.com/Kentico/delivery-sdk-net/issues/152

PetrSvirak:
> > If we agree on a new major, though, I would continue with refactoring a bit further and removed `ResolvedContentItemData` completely (definitely a breaking change, of course) and maybe even tried to figure out an even simpler way to create simple strongly typed inline resolvers (this would require a bit more validation – maybe [switch with pattern matching](https://docs.microsoft.com/en-us/dotnet/csharp/pattern-matching#using-pattern-matching-switch-statements) is our best bet anyway).
> 
> @PetrSvirak , I would personally hate to see the use of switch pattern here. If you had a switch, you would still had to cast the item to your proper type in order to work with it anyway. I feel that using `is` is superior and would leave it at that.

Thanks for your opinion on `switch`. I personally love how the newish pattern matching shortens the code and makes it less cluttered (but is just my opinion).

<details>
  <summary>
    I took my liberty to rewrite the code you provided in #153 in the fashion I have in mind 
    (since code says is it better than words, right?)
</summary>
<p>

```csharp
{
	{
		switch{
			case InfoBox infoBox:
				return $"<div>{infoBox.Content}</div>";

			case CodeBlock codeBlock:
				return
					$"<pre><code class=\"language-{codeBlock.AvailableLanguagesLanguage?.FirstOrDefault()?.Codename.ToLower().Trim()}\">" +
					$"{System.Web.HttpUtility.HtmlEncode(codeBlock.Code)?.Trim()}" +
					$"</code></pre>";

			default:
		}
	}
}
```
As you see, there is no extra casting and all variables are just as scoped as the ones in `if` statements, there is less chaining and clear vertical alignment that helps following the code (IMO). But it is definitely a matter of taste, at least to some extent, and I'm the last person to seek a general best or worst way to write a piece of code.

</p>
</details>

Enngage:
Yeah, this would work too :) I initially though it would work a bit differently, but as you say - it is mainly a preference thing. I like to avoid using switch wherever I can, but someone else might like it :)

PR #156, opened 2019-03-07
Get rid of code-first naming

### LINKED ISSUE #152: Code First approach
The [wiki page describing how to work with strong typed models](https://github.com/Kentico/delivery-sdk-net/wiki/Working-with-Strongly-Typed-Models-(aka-Code-First-Approach)), states that it's also known as the "code first approach". Reading this made me think that the models you define in code are leading (just like Entity Framework's code first). This however doesn't seem to be the case since the content types always have to be defined via the Kentico Cloud UI.

I would suggest changing the title because I think it's not really code-first. It would be a nice feature to create the content types based on the code though!

matus12:
### Motivation

Which issue does this fix? Fixes #152 

### Checklist

- [x] Tests are passing

### How to test


petrsvihlik:
nice!

PR #166, opened 2019-09-05
Feature/kcl 1219 support for items feed

sutrkiller:
### Motivation

Support for a new Delivey API endpoint /items-feed

### Checklist

- [ ] Tests are passing

### How to test

Automated tests should do it.


KenticoCloud.Delivery/StrongTyping/DeliveryItemsFeed.cs
```diff
@@ -0,0 +1,79 @@
...
+                _lastResponse = await _getFeedResponseAsync(_lastResponse?.ContinuationToken);
+            }
+
+            return _lastResponse;
+        }
+
+        /// <summary>
+        /// Returns an enumerator that iterates through the collection of content items from all batches.
+        /// </summary>
+        public IEnumerator<T> GetEnumerator()
```
Enngage: @sutrkiller, I'm not familiar with the new endpoint, but isn't this a bit dangerous? I mean - how many API calls can you 'accidentally' make by iterating through all batches? Say you have 100k content items, how would this work?

KenticoCloud.Delivery/StrongTyping/DeliveryItemsFeed.cs
```diff
@@ -0,0 +1,79 @@
...
+    {
+        internal delegate Task<DeliveryItemsFeedResponse<T>> GetFeedResponse(string continuationToken);
+
+        private DeliveryItemsFeedResponse<T> _lastResponse;
+        private readonly GetFeedResponse _getFeedResponseAsync;
+
+        /// <summary>
+        /// Indicates whether there are more batches to fetch.
+        /// </summary>
+        public bool HasMoreResults => _lastResponse == null || !string.IsNullOrEmpty(_lastResponse.ContinuationToken);
```
Enngage: @sutrkiller HasMoreResults is true if last response is null, but that doesn't necessarily means there are more results. I would consider adding new property to indicate if some batch was already fetched and 'HasMoreResults' should be equal to true only if there is valid continuationToken

KenticoCloud.Delivery/Responses/DeliveryItemsFeed.cs
```diff
@@ -0,0 +1,78 @@
...
+    {
+        internal delegate Task<DeliveryItemsFeedResponse> GetFeedResponse(string continuationToken);
+
+        private DeliveryItemsFeedResponse _lastResponse;
+        private readonly GetFeedResponse _getFeedResponseAsync;
+
+        /// <summary>
+        /// Indicates whether there are more batches to fetch.
+        /// </summary>
+        public bool HasMoreResults => _lastResponse == null || !string.IsNullOrEmpty(_lastResponse.ContinuationToken);
```
Enngage: Same as other comment below.

KenticoCloud.Delivery/StrongTyping/DeliveryItemsFeed.cs
sutrkiller: Yes, it is a little dangerous as it can end up in more API calls. This endpoint does not return modular content and will be used only by a fraction of users. This is necessary for current users who will need to transfer to this endpoint with a least amount of changes to their codebase.
Compare: 
`foreach (var item in await _client.GetItemsAsync())`
vs 
`foreach (var item in _client.GetItemsFeed()`

KenticoCloud.Delivery/StrongTyping/DeliveryItemsFeed.cs
sutrkiller: I agree that the name 'HasMoreResult' might be misleading but a call to FetchNextBatch would be necessary anyway. Again, conveniance was a main reason for this I cannot come up with any use case for the new proposed property as the expected usage of Feed is something like:
```
while (feed.HasMoreResults) 
{
    var nextBatch = await feed.FetchNextBatchAsync();
   // do something with items
}
```
or do-while alternatively...

sutrkiller:
Please do not merge yet. It should be delayed until beta of 11 is released.

KenticoCloud.Delivery/StrongTyping/DeliveryItemsFeed.cs
Enngage: I'm honestly not sure this is a good idea. Saying that only a few people will use it (for now) is not very satisfactory. People will eventually use it in some way or the other. I'm not that worried about the size of the response, but by the number of queries it can make. 

Image developer not familiar of what this does - they call this method on their dev environment (where they have only a handful of items) and everything works nicely, so they move their code to prod and everything fails because now it doesn't make 1-2 HTTP requests, it makes 1 000. Depending on the implementation you can pretty much kill your server for everyone. 

If we want to use this, we should be **very explicit** of what it does. I mean, even if you had to make 'just' 20 requests to get all items, it could still be a relatively long wait - 20 * average response time of 50ms = 1 sec - just because devs don't know what this does under the hood.

KenticoCloud.Delivery/StrongTyping/DeliveryItemsFeed.cs
Enngage: Yeah, I guess the example you posted is reasonable and easy to use so we can leave this as it is.

KenticoCloud.Delivery/StrongTyping/DeliveryItemsFeed.cs
sutrkiller: We decided to remove the potentially dangerous code.

PR #170, opened 2019-09-24
Detect stale content

sutrkiller:
### Motivation

To improve response times Delivery API might provide stale content. Stale means that there is a more recent version, but it will become available later. This information is important for developers who implement caching because stale content should have a short TTL.

Developers can inspect the value of the **X-Stale-Content** response header. **1** indicates that content is stale. With SDK developers can also inspect the value of the **HasStaleContent** property of response objects.

This feature introduced a breaking change. Documentation will be updated independently.

### Checklist

- [x] Tests are passing

### How to test

There is no need for manual testing.


sutrkiller:
This is a revert of once reverted changes of detect stale content (see closed PRs).

PR #172, opened 2019-09-25
GH Issue 171: asset missing height width properties

### LINKED ISSUE #171: Asset does not have Height or Width values
The API shows `height` and `width` properties in any linked assets coming back in the JSON, but the `Asset` class provided by `KenticoCloud.Delivery` does not have `Height` or `Width` properties.

We are also using the model generator which regenerates the models when we execute it, so would override our implementation of any `CustomAsset` implementation in our generated models.

OnyxPrime:
### Motivation

Which issue does this fix? Fixes #171 

### Checklist

- [x] Tests are passing


OnyxPrime:
Hey @makma, thank you for catching that. I've added the properties to the constructor and fixed the spacing. Oddly though, the spacing looked correct in VS, VSCode and Notepad++, so I ended up fixing in browser.

PR #191, opened 2020-02-26
Feature/cache manager

tomasjurasek:
### Motivation

Which issue does this fix? Fixes #`issue number`

### Checklist

- [ ] Tests are passing

### How to test

### DIFF
--- a/Kentico.Kontent.Delivery.Abstractions/IDeliveryCacheManager.cs
+++ b/Kentico.Kontent.Delivery.Abstractions/IDeliveryCacheManager.cs
@@ -5,11 +5,42 @@ using System.Threading.Tasks;
 
 namespace Kentico.Kontent.Delivery.Abstractions
 {
+    /// <summary>
+    /// Cache responses against the Kentico Kontent Delivery API.
+    /// </summary>
     public interface IDeliveryCacheManager
     {
+        /// <summary>
+        /// Returns or Adds data to the cache
+        /// </summary>
+        /// <typeparam name="T">A generic type</typeparam>
+        /// <param name="key">A cache key</param>
+        /// <param name="valueFactory">A factory which returns a data</param>
+        /// <param name="shouldCache"></param>
+        /// <param name="dependenciesFactory"></param>
+        /// <returns>The data of a generic type</returns>
         Task<T> GetOrAddAsync<T>(string key, Func<Task<T>> valueFactory, Func<T, bool> shouldCache = null, Func<T, IEnumerable<string>> dependenciesFactory = null);
+
+        /// <summary>
+        /// Tries to return a data
+        /// </summary>
+        /// <typeparam name="T">Generic type</typeparam>
+        /// <param name="key">A cache key</param>
+        /// <param name="value">Returns data in out parameter if are there.</param>
+        /// <returns>Returns true or false.</returns>
         Task<bool> TryGetAsync<T>(string key, out T value);
+
+        /// <summary>
+        /// Invalidates data by the key
+        /// </summary>
+        /// <param name="key">A cache key</param>
+        /// <returns></returns>
         Task InvalidateDependencyAsync(string key);
+
+        /// <summary>
+        /// Clears cache
+        /// </summary>
+        /// <returns></returns>
         Task ClearAsync();
     }
 }
--- a/Kentico.Kontent.Delivery.Abstractions/IDeliveryCacheManager.cs
+++ b/Kentico.Kontent.Delivery.Abstractions/IDeliveryCacheManager.cs
@@ -1,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Kentico.Kontent.Delivery.Abstractions
+{
+    public interface IDeliveryCacheManager
+    {
+        Task<T> GetOrAddAsync<T>(string key, Func<Task<T>> valueFactory, Func<T, bool> shouldCache = null, Func<T, IEnumerable<string>> dependenciesFactory = null);
+        Task<bool> TryGetAsync<T>(string key, out T value);
+        Task InvalidateDependencyAsync(string key);
+        Task ClearAsync();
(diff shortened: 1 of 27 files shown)


Kentico.Kontent.Delivery/DeliveryClient.cs
```diff
@@ -27,6 +28,7 @@ internal sealed class DeliveryClient : IDeliveryClient
         internal readonly IPropertyMapper PropertyMapper;
         internal readonly IRetryPolicyProvider RetryPolicyProvider;
         internal readonly IDeliveryHttpClient DeliveryHttpClient;
+        internal readonly IDeliveryCacheManager DeliveryCacheManager;
```
petrsvihlik: I would strongly recommend sticking to the Decorator pattern that was introduced in https://github.com/Kentico/kontent-boilerplate-net/blob/master/src/content/Kentico.Kontent.Boilerplate/Caching/Webhooks/CachingDeliveryClient.cs

It'll help us keep the responsibilities of classes separate.

Here's a suggested way of doing it with .NET Core's DI: https://greatrexpectations.com/2018/10/25/decorators-in-net-core-with-dependency-injection

Kentico.Kontent.Delivery/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -186,6 +188,8 @@ private static IServiceCollection RegisterDependencies(this IServiceCollection s
             services.TryAddSingleton<IRetryPolicyProvider, DefaultRetryPolicyProvider>();
             services.TryAddSingleton<IDeliveryClient, DeliveryClient>();
             services.TryAddSingleton<IDeliveryClientFactory, DeliveryClientFactory>();
+            services.TryAddSingleton<IDeliveryCacheManager, DeliveryCacheManager>();
+            services.TryAddSingleton<IMemoryCache, MemoryCache>();
```
petrsvihlik: If we use decorators, we can avoid registering default caching implementations every time. We can do that only when needed - in `WithDeliveryCacheManager()`.

Kentico.Kontent.Delivery/DeliveryClient.cs
```diff
@@ -141,10 +164,18 @@ public async Task<DeliveryItemResponse> GetItemAsync(string codename, IEnumerabl
...
+                CacheHelpers.GetItemKey(codename, parameters),
+                async () =>
+                {
+                    var endpointUrl = UrlBuilder.GetItemUrl(codename, parameters);
+                    var response = await GetDeliverResponseAsync(endpointUrl);
+
+                    return new DeliveryItemResponse(response, ModelProvider, ContentLinkResolver);
+                },
+                response => response != null,
+                CacheHelpers.GetItemDependencies);
```
petrsvihlik: If we use decorators, we can avoid performing costly operations such as `CacheHelpers.Get*Dependencies` every time, even if the caching is not used.

Kentico.Kontent.Delivery.Abstractions/IDeliveryCacheManager.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Threading.Tasks;
+
+namespace Kentico.Kontent.Delivery.Abstractions
+{
+    public interface IDeliveryCacheManager
+    {
+        Task<T> GetOrAddAsync<T>(string key, Func<Task<T>> valueFactory, Func<T, bool> shouldCache = null, Func<T, IEnumerable<string>> dependenciesFactory = null);
+        Task<bool> TryGetAsync<T>(string key, out T value);
+        Task InvalidateDependencyAsync(string key);
+        Task ClearAsync();
```
petrsvihlik: Missing comments for publicly accessible members.

Kentico.Kontent.Delivery/Cache/DeliveryCacheManager.cs
```diff
@@ -0,0 +1,161 @@
...
+using System.Threading.Tasks;
+
+namespace Kentico.Kontent.Delivery.Cache
+{
+    /// <summary>
+    /// Cache responses against the Kentico Kontent Delivery API.
+    /// </summary>
+    public sealed class DeliveryCacheManager : IDeliveryCacheManager
+    {
+        private readonly IMemoryCache _memoryCache;
```
petrsvihlik: Just a note: If we'll want to support [distributed caching](https://docs.microsoft.com/en-us/aspnet/core/performance/caching/distributed?view=aspnetcore-3.1) we'll need to create a different implementation of IDeliveryCacheManager that will depend on [`IDistributedCache`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.caching.distributed.idistributedcache?view=dotnet-plat-ext-3.1) rather than [`IMemoryCache`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.caching.memory.imemorycache?view=dotnet-plat-ext-3.1) 

ping @xantari

petrsvihlik: https://github.com/Kentico/kontent-boilerplate-net/issues/90

Kentico.Kontent.Delivery/DeliveryClient.cs
tomasjurasek: In the current version of the SDK is not possible to use this patter.

We installed all default dependency in the ServiceCollection extensions - you can rewrite them in separate registration. If we should use the Decorator pattern we don't very clear way how to do it with the actual architecture of the SDK.

Because things like caching and policy we can set up in the Configuration - Options and we cannot set up a dependency, which depends on this configuration.

Now we can control behavior by the Configuration - Options just in implementations classes like IDeliveryClient,..

Kentico.Kontent.Delivery/DeliveryClient.cs
tomasjurasek: If caching is not enabled these methods are not called. It's a Func which is resolved only in CacheManager.

Issue #217, opened 2020-05-05
Refactor responses


petrsvihlik:
Expose the raw response
- refactor the ApiResponse 
  - replace public JObject Content with string
- refactor AbstractResponse
  - make ApiResponse public
- create interfaces

Get rid of ContentItem
- remove Kentico.Kontent.Delivery.Abstractions.ContentItem
- remove related methods from the `IDeliveryClient`
- make sure existing scenarios can be worked around with strongly-typed responses

Issue #196, opened 2020-02-25
Add support for IDistributedCache to the SDK


xantari:
### Motivation
Currently, the SDK provides a CacheManager that uses `IMemoryCache`.

In order to support more advanced scenarios utilizing commonly used distributed caching libraries such as:

- Microsoft.Extensions.Caching.Distributed.MemoryDistributedCache
- Microsoft.Extensions.Caching.Redis.RedisCache
- Microsoft.Extensions.Caching.SqlServer.SqlServerCache
- Microsoft.Extensions.Caching.StackExchangeRedis.RedisCache

we'd like to introduce support for the [`IDistributedCache` interface](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.caching.distributed.idistributedcache?view=dotnet-plat-ext-3.1).

### Proposed solution
There are several approaches that we can take:
1. Create a new implementation if the [`IDeliveryCacheManager`](https://github.com/Kentico/kontent-delivery-sdk-net/blob/master/Kentico.Kontent.Delivery.Abstractions/IDeliveryCacheManager.cs) that would be able to serialize the `Delivery*Response` objects. Depending on whether the implementation will be [`byte[]`-based or `string`-based](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.caching.distributed.idistributedcache?view=dotnet-plat-ext-3.1), it will require some changes in the `Delivery*Response` and underlying objects (such as Taxonomy) in terms of decorating them with certain attributes, implementing serializers, etc.
If we choose to go this way, I'd vote for:
- moving all the logic from the `Delivery*Response` objects to the `DeliveryClient` (everything that's related to `IModelProvider` or `Newtonsoft.Json`)
- removing `Kentico.Kontent.Delivery.Abstractions`'s dependency on `Newtonsoft.Json.Linq`
- getting rid of `dynamic` properties - I think we can replace them with strongly-typed collections of objects (not sure why we haven't done so yet)
- make all the models POCO with public constructors
This approach is best performance-wise as it stores the objects after the strong typing facilitated by `IModelProvider`. And also goes in line with the caching model we've laid out with the introduction of the `IDeliveryCacheManager`.
2. Implement an HTTP response-level caching. This can be achieved e.g. using Polly
- https://nodogmablog.bryanhogan.net/2018/11/caching-in-polly-6-and-the-httpclientfactory/
- https://github.com/App-vNext/Polly/wiki/Cache
[...]

Issue #61, opened 2017-07-17
Spike: Support interfaces as generic parameters of collections


petrsvihlik:
### Motivation

In [strongly typed models](https://github.com/Kentico/delivery-sdk-net#strongly-typed-responses), this is supported:

`IEnumerable<MyType> CollectionOfMyTypes { get; set; }`

and this is not:

`IEnumerable<IMyType> CollectionOfMyInterfaces { get; set; }`

This applies both to externally defined types and types defined within the SDK (`Asset`, `TaxonomyTerm`).

We should give developers the freedom to unwind from specific classes and work with Kontent data through interfaces. For example, [defining shadow properties](https://github.com/Kentico/kontent-delivery-sdk-net/wiki/Strong-Types-Explained-%E2%80%93-Model-Inheritance#define-shadow-properties-aside-from-existing-ones) wouldn't be necessary anymore.

### Expected spike outcome

Propose a way of defining translations between interfaces and concrete types.

These translations would be used in [CodeFirstModelProvider](https://github.com/Kentico/kontent-delivery-sdk-net/blob/cc925759439ea591e1f019d729653b841a793fdb/Kentico.Kontent.Delivery/StronglyTyping/ModelProvider.cs) wherever `Activator.CreateInstance(...);` is called.

There would be default translations for `Asset`, `TaxonomyTerm` and other types defined in the SDK to maintain backwards compatibility.

As for **how** we could allow the `CodeFirstModelProvider` to translate interfaces to objects, maybe we can leverage the good old IoC pattern. A brief spike revealed that 3 out of 4 popular IoC containers for .NET Core use the [IServiceCollection ](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.iservicecollection) abstraction to work with the services and their implementations.

Namely:

* [Lamar](https://jasperfx.github.io/lamar/documentation/ioc/aspnetcore/)'s [ServiceRegistry](https://github.com/JasperFx/lamar/blob/master/src/Lamar/ServiceRegistry.cs#L40) class implements `IServiceCollection`.
* [AutoFac](http://autofac.readthedocs.io/en/latest/integration/netcore.html) works directly with [ServiceCollection](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.servicecollection), which is a default implementation of `IServiceCollection`.
[...]

PR #218, opened 2020-05-30
Feature/refactoring models

### LINKED ISSUE #216: Change IDeliveryClient to make faking responses simpler
### Motivation

In order to fake responses from the `IDeliveryClient` you currently have to mock the `HttpClient` and set up the desired json as set out in the [Faking responses wiki page](https://github.com/Kentico/kontent-delivery-sdk-net/wiki/Faking-responses). This feels like more effort than should be required, and involves too much Kontent infrastructure in my unit tests. Ideally you should be able to do something like this:

```csharp
var deliveryClient = A.Fake<IDeliveryClient>();
var deliveryItemResponse = A.Fake<DeliveryItemResponse<Article>>();
A.CallTo(deliveryClient).WithReturnType<DeliveryItemResponse<Article>>().Returns(deliveryItemResponse);
```
However, this is not possible because the constructor for the `DeliveryItemResponse` is internal, so the above will throw an exception.

### Proposed solution

A quick solution would be to make the relevant constructors public but that feels like a bad design choice. I propose the creation of an `IDeliveryItemResponse` interface which should be used as the return type for the `IDeliveryClient` methods.

petrsvihlik:
### Motivation

Which issue does this fix? Fixes #216, #217, and partially #61

The changes should reduce maintenance, promote good API design, and serve as a foundation for #196 & other improvements. 

### Changes & Notes for the reviewer

The changes include:
- getting rid of `dynamic` fields -> replaced with strongly-typed `IReadOnlyList<object> LinkedItems { get; }` with models resolved
- getting rid of weakly-typed `ContentItem` & related code (they didn't support some of the latest features and were really difficult to maintain) - there is no replacement, the consumer needs to call `JObject.Parse(ApiResponse.Content)` and retrieve/transform the data manually
- simplifying the `IDeliveryClient` interface to a bare minimum (from 25 methods to 8)
-- some of them are still available as `DeliveryClientExtensions`
- removing unnecessary dependencies from the `Abstractions` library (like `Newtonsoft.Json`)
- removing unnecessary dependencies from the `Caching` library (it's now only dependent on Abstractions)
- namespace cleanup (classes grouped by their purpose/semantics...no longer by whether they are interfaces, abstract classes, middleware, etc.)
- `IPropertyValueConverter` is more specific now - uses `IContentElement elementData` instead of JToken
- simplified interface structure around the `IContentLinkUrlResolver` -> removed the parent interface
- switching to async alternatives where possible (unit tests, etc)

### Testing package
https://www.nuget.org/packages/Kentico.Kontent.Delivery/14.0.0-beta1

### Current dependency diagram
image

### Checklist

- [x] Tests are passing


petrsvihlik:
@xantari @robertgregorywest could you guys please take a look and verify that the new API works in your scenarios?

I'm mostly concerned about:
- the members removed from the `IDeliveryClient` interface (JObject & ContentItem-based responses)
- getting rid of the `dynamic` field in the responses

It'd also be great if you could try the beta package. (There will be tons of breaking changes, I'm sure but most of them is just a namespace change.)

Any help appreciated!

xantari:
@petrsvihlik Sure! Might take me a day or two before I can get back to you on this.

robertgregorywest:
@petrsvihlik no problem, I'll take a look now and install that beta.

Kentico.Kontent.Delivery.Abstractions/IDeliveryClientExtensions.cs
```diff
@@ -0,0 +1,72 @@
...
+        /// Gets a strongly typed content item by its codename. By default, retrieves one level of linked items.
+        /// </summary>
+        /// <typeparam name="T">Type of the model. (Or <see cref="object"/> if the return type is not yet known.)</typeparam>
+        /// <param name="client">An instance of the <see cref="IDeliveryClient"/></param>
+        /// <param name="codename">The codename of a content item.</param>
+        /// <param name="parameters">An array that contains zero or more query parameters, for example, for projection or setting the depth of linked items.</param>
+        /// <returns>The <see cref="IDeliveryItemResponse{T}"/> instance that contains the content item with the specified codename.</returns>
+        public static async Task<IDeliveryItemResponse<T>> GetItemAsync<T>(this IDeliveryClient client, string codename, params IQueryParameter[] parameters)
+        {
+            return await client.GetItemAsync<T>(codename, parameters);
```
seangwright, line 23: You can [elide these calls](https://blog.stephencleary.com/2016/12/eliding-async-await.html) since result of the task is not needed after the call no operations are performed before the call.

petrsvihlik: true. great article, I never realized that :)
I'll incorporate the change

robertgregorywest:
The beta works fine once I dealt with the namespace changes. I was using the Kentico.Kontent.AspNetCore NuGet package which doesn't have a prerelease for this beta, so just added a local copy of the tag helper to deal with the namespace change.

Testing is much easier now that the client returns an interface, so all good from my perspective.

Thanks,

Rob

petrsvihlik:
perfect! thanks very much! let' wait for what @xantari has to say about the changes.

I asked @tomasjurasek for a review from our side and I'd like to invite everyone who's spent some considerable time with the SDK in the past... @Enngage or @alesk-kentico come to mind.

Kentico.Kontent.Delivery.Abstractions/IDeliveryClientExtensions.cs
petrsvihlik: resolved in c993a9bd22f337b003b5ffcd968e2017a1827e0b

xantari:
I got nearly through all the namespace changes, most of them were in the generated models (which I looked for an updated KontentGenerator but could only find 5.0.1, i'm guessing that update will be coming soon, so that the updated models will have the new namespaces already included).

What I need help with is the below bit of code. I read your notes, but not sure how best to handle this.

I was doing querys on system.type, for the 3 top level things that could be distinct "pages" in our website. All other content types are just items you can put into one of the three types of pages.

Here is the screenshot of the errors:

image

Here is the relevant code of that class:

```
     public static async Task<List<ControllerMapping>> GetControllerMappings(this IDeliveryClient deliveryClient, ITaxonomyGroup sitemapTaxonomy)
        {
            var parameters = new List<IQueryParameter>
                {
                    new DepthParameter(5),
                    new InFilter("system.type", LandingPage.Codename, ContentPage.Codename, NewsArticle.Codename), //This is a list of all the "page types" in kentico
                };
            var response = await deliveryClient.GetItemsAsync(parameters);

            var controllerMappings = new List<ControllerMapping>();

            foreach (var page in response.Items)
            {
                controllerMappings.Add(GetControllerMapping(page, sitemapTaxonomy));
            }
            return controllerMappings;
        }

        private static ControllerMapping GetControllerMapping(ContentItem item, TaxonomyGroup sitemapTaxonomy)
        {
            ControllerMapping mapping = new ControllerMapping();

            if (item.System.Type == ContentPage.Codename)
            {
                var page = item.CastTo<ContentPage>();
                mapping.FullUrl = item.GetPageUrl(sitemapTaxonomy);
                mapping.PageCodeName = item.System.Codename;
                mapping.PageUrlSlug = page.UrlSlug;
                mapping.ControllerName = "ContentPage";
            }

            if (item.System.Type == LandingPage.Codename)
            {
                var page = item.CastTo<LandingPage>();

                mapping.FullUrl = item.GetPageUrl(sitemapTaxonomy);
                mapping.PageCodeName = item.System.Codename;
                mapping.PageUrlSlug = page.UrlSlug;
                mapping.ControllerName = "LandingPage";
            }

            if (item.System.Type == NewsArticle.Codename)
            {
                var page = item.CastTo<NewsArticle>();

                mapping.FullUrl = item.GetPageUrl(sitemapTaxonomy);
                mapping.PageCodeName = item.System.Codename;
                mapping.PageUrlSlug = page.UrlSlug;
                mapping.ControllerName = "News";
            }

            mapping.FullUrl = mapping.FullUrl.TrimEnd('/');

            return mapping;
        }
```

Can you query by system.type anymore like that?

petrsvihlik:
@xantari gotcha. there's a thing that we call ["runtime typing"](https://github.com/Kentico/kontent-delivery-sdk-net/wiki/Strong-Types-Explained-%E2%80%93-Runtime-Typing#how-to-use-runtime-typing) that is probably an answer to your question. 

so the code would look like this in your case:
```
var parameters = new List<IQueryParameter>
		{
			new DepthParameter(5),
			new InFilter("system.type", LandingPage.Codename, ContentPage.Codename, NewsArticle.Codename), //This is a list of all the "page types" in kentico
		};
	var response = await deliveryClient.GetItemsAsync<object>(parameters);

            foreach (var item in response.Items)
            {
                switch (item)
                {
                    case LandingPage _:
                        {
                            var page = item as LandingPage;
                            var xyz = page.PropertyA;
                            break;
                        }

                    case ContentPage _:
                        {
                            var page = item as ContentPage;
                            var xyz = page.PropertyB;
                            break;
                        }
                }
            }
```
or you can query it like:
```
var landingPages = response.Items.OfType<LandingPage>();
```
additionally, you can retrieve the type name as string:
```
var ctp = new CustomTypeProvider();
var strType = ctp.GetCodename(item.GetType());
```


and yes, the code generator will be updated soon as well.

One more question - when you're retrieving collections - is it sufficient to have just the type information available? Would it be handy in any way to have a way of retrieving `ContentItemSystemAttributes` before you cast the item from object to a specific strong type? (E.g. to be able to query the items by language for instance? Or does such filtering happen rather on the REST API level?)

xantari:
So far everything is working OK with the upgrade to beta1. It would be handy to be able to view the ContentItemSystemAttributes. That gives flexibility of doing the filtering on the local (cached) side in addition to being able to do it from the REST API level.

i'll be pushing out the site with beta1 to our test environment hopefully in a few days and will let you know if we encounter anything else.

petrsvihlik:
@xantari that's perfect! 

here's the generator: https://www.nuget.org/packages/Kentico.Kontent.ModelGenerator/6.0.0-beta2
(works with the [beta3](https://www.nuget.org/packages/Kentico.Kontent.Delivery/14.0.0-beta3) SDK) 

I'll take a look at the `ContentItemSystemAttributes`

petrsvihlik:
the most recent changes are:
-  that Assets, Taxonomy groups, and MultipleChoice elements are interfaces (the generator outputs e.g. `IEnumerable<IAsset>` and the SDK fills it with a specific implementation)
- and that there is no namespace nesting in the Abstractions library (it would generate too much unnecessary code)

Enngage:
We are using **Linked items** element with type contraints for about a dozen of **Content types**. In SDK we extracted an interface for these types and use it like:

```
public IEnumerable<IBaseComponent> Components { get; set; }
```

However, this fails because SDK cannot create an instance of `IBaseComponent` type. This worked in previous versions of SDK and particular items in `Components`  were mapped to appropriate types using registered type resolvers. 

The workaround is fairly simple though as all we need to do is `Cast` the objects:

```
public IEnumerable<IBaseComponent> MappedComponents => Components.Cast<IBaseComponent>();
```

petrsvihlik:
@Enngage this has probably been removed accidentally some time ago. fixed in d79c43379d3c33bd792585d888ca837b7f98b816

PR #221, opened 2020-07-10
Add a new AddDeliveryClient method

tomasjurasek:
### Motivation

Which issue does this fix? Fixes #`issue number`

### Checklist

- [x] Tests are passing

### How to test

### DIFF
--- a/Kentico.Kontent.Delivery/Extensions/ServiceCollectionExtensions.cs
+++ b/Kentico.Kontent.Delivery/Extensions/ServiceCollectionExtensions.cs
@@ -51,7 +51,6 @@ namespace Kentico.Kontent.Delivery.Extensions
                 .RegisterFactoryDependencies();
         }
 
-
         /// <summary>
         /// Registers a delegate that will be used to configure a named <see cref="IDeliveryClient"/> via <see cref="IDeliveryClientFactory"/>
         /// </summary>
--- a/Kentico.Kontent.Delivery/Extensions/ServiceCollectionExtensions.cs
+++ b/Kentico.Kontent.Delivery/Extensions/ServiceCollectionExtensions.cs
@@ -25,6 +25,38 @@ namespace Kentico.Kontent.Delivery.Extensions
         /// </summary>
         ///<param name="name">The name of the client configuration</param>
         /// <param name="services">A <see cref="ServiceCollection"/> instance for registering and resolving dependencies.</param>
+        /// <param name="buildDeliveryClient">A function that returns a valid instance of <see cref="IDeliveryClient"/>.</param>
+        /// <returns>The <paramref name="services"/> instance with <see cref="IDeliveryClient"/> registered in it</returns>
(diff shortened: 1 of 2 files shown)


Kentico.Kontent.Delivery/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -20,6 +20,38 @@ namespace Kentico.Kontent.Delivery.Extensions
...
     public static class ServiceCollectionExtensions
     {
+        /// <summary>
+        /// Registers a delegate that will be used to configure a named <see cref="IDeliveryClient"/> via <see cref="IDeliveryClientFactory"/>
+        /// </summary>
+        ///<param name="name">The name of the client configuration</param>
+        /// <param name="services">A <see cref="ServiceCollection"/> instance for registering and resolving dependencies.</param>
+        /// <param name="buildDeliveryClient">A function that returns a valid instance of <see cref="IDeliveryClient"/>.</param>
+        /// <returns>The <paramref name="services"/> instance with <see cref="IDeliveryClient"/> registered in it</returns>
+        public static IServiceCollection AddDeliveryClient(this IServiceCollection services, string name, Func<IDeliveryClient> buildDeliveryClient)
```
PetrSvirak, line 30: We used `Func<IDeliveryOptionsBuilder, DeliveryOptions>` to make the building more natural.

Providing a builder sets expectations - you might not know there is a builder when you first try to call the method and nobody forces you to actually use the builder if you don't want to - method just gives you a little nudge. 

Especially for `Options`, it saves knowing of and calling `DeliveryOptionsBuilder.CreateInstance();` since API performs a little code like this for your:
```csharp
            var builder = DeliveryOptionsBuilder.CreateInstance();
            var options = buildDeliveryOptions(builder);
```

Would you considered updating the method header this way?

petrsvihlik: not exactly sure what you are trying to say. do you think we should change the signature to expect an `IDeliveryClientBuilder`? would you mind sharing a piece of code to demonstrate it?

PetrSvirak: Yeah, sure. Seems I did not manage to express myself properly, sorry.

What I lacked when reading the method signature was something like:
```csharp
public static IServiceCollection AddDeliveryClient(this IServiceCollection services, string name, Func<IDeliveryClientBuilder, IDeliveryClient> buildDeliveryClient)
```

The same concept that is used with options builder for example in the following overload:
```csharp
public static IServiceCollection AddDeliveryClient(this IServiceCollection services, string name, Func<IDeliveryOptionsBuilder, DeliveryOptions> buildDeliveryOptions)
```

Tomas seems to already incorporate the change 👍

petrsvihlik: thanks for the clarification! I thought so :)

Issue #222, opened 2020-08-03
Add performance tests


petrsvihlik:
### Motivation
In order to be able to track and improve the number of requests the Delivery SDK is able to handle, we should add performance tests.

We should add both CLR-level benchmarks and a sample load-test web service to test the SDK in real-world conditions (multithreading, garbage collection, etc.)

### Proposed solution

1) Add benchmarks - https://github.com/dotnet/BenchmarkDotNet
    - looking at Serilog, it should be possible to [run the benchmarks with xunit](https://github.com/serilog/serilog/blob/master/test/Serilog.PerformanceTests/Harness.cs) when [`shadowCopy` is turned off](https://github.com/serilog/serilog/blob/master/test/Serilog.PerformanceTests/xunit.runner.json). we need to test it and possibly ([answer this question](https://stackoverflow.com/questions/60828785/running-benchmarkdotnet-within-xunit))
    - another great example is [autofac](https://github.com/autofac/Autofac/tree/master/bench/Autofac.Benchmarks)
    - or add a testing console app similarly to [newtonsoft.json](https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json.TestConsole/Program.cs) + [benchmarks in newtonsoft.json](https://github.com/JamesNK/Newtonsoft.Json/tree/master/Src/Newtonsoft.Json.Tests/Benchmarks)
    - alternatively, we can use [xunit-performance](https://github.com/Microsoft/xunit-performance) add-on.

2) Add a load test project
    - Since [load tests were deprecated in VS](https://devblogs.microsoft.com/devops/cloud-based-load-testing-service-eol/) we'll have to rely on a different tool. A good one would be [bombardier](https://github.com/codesenberg/bombardier). As described in this video: https://www.youtube.com/watch?v=jn54CjePzs0

### Other resources
https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-5/?ocid=3017123&MC=.NET&MC=CSHARP&MC=Vstudio

petrsvihlik:
the first batch of tests was added in #226

Issue #213, opened 2020-04-22
IContentLinkUrlResolver not async friendly


xantari:
### Brief bug description

I wasn't sure if I should file this as a bug report or a feature request.

I have switched to your latest version of the Kentico.Kontent.Delivery.Caching nuget (13.0.1) and removed all the Caching client code we got from the boiler plate example now that you built in this functionality.

The problem is that we store additional items into the IMemoryCache to store the generated SEO URL's that are based off of the web pages taxonomy selection in Kentico Kontent.

So our IContentLinkUrlResolver needs the ability to lookup that information from the cache.

However, the IContentLinkUrlResolver is not async friendly and you have switched your IDeliveryCacheManager to only have async available commands where as the previous boiler plate had them as non-async.

Here is what our IContentLinkUrlResolver looks like today. Notice I already changed the constructor to get the injected IDeliveryCacheManager from your new Kentico.Kontent.Delivery.Caching nuget.

```
    public class CustomContentLinkUrlResolver : IContentLinkUrlResolver
    {
        private IDeliveryCacheManager _cachingClient;

        public CustomContentLinkUrlResolver(IDeliveryCacheManager cacheClient)
        {
            _cachingClient = cacheClient;
        }

        /// <summary>
        /// Resolves the link URL.
        /// </summary>
        /// <param name="link">The link.</param>
        /// <returns>A relative URL to the page where the content is displayed</returns>
        public string ResolveLinkUrl(ContentLink link)
        {
            List<ControllerMapping> mappings;
            bool result = _cachingClient.TryGet<List<ControllerMapping>>(Constants.ControllerMappingCacheKey, out mappings);

            if (mappings != null && mappings.Count() > 0)
            {
                var contentItem = mappings.FirstOrDefault(p => p.PageCodeName == link.Codename);
                return contentItem.FullUrl;
            }
            return string.Empty;
        }

        /// <summary>
        /// Resolves the broken link URL.
        /// </summary>
        /// <returns>A relative URL to the site's 404 page</returns>
        public string ResolveBrokenLinkUrl()
        {
            // Resolves URLs to unavailable content items
            return "/404";
        }
    }
```

However the issue is on this line:

```
            bool result = _cachingClient.TryGet<List<ControllerMapping>>(Constants.ControllerMappingCacheKey, out mappings);
```

There is no non-async version of TryGet available anymore.

[...]

PR #226, opened 2020-08-13
Distributed cache + Async API

petrsvihlik:
### Motivation

Fixes 
- #213  - IContentLinkUrlResolver not async friendly (recommended reading)
- #196 - Add support for IDistributedCache to the SDK - approach no. 1 (recommended reading)
- #222 - Add performance tests

Follows up on #218 (also a recommended reading for understanding the context)

### Checklist

- [x] Tests are passing
- [ ] Docs have been updated (if applicable) -> will be

### Breaking changes, features & upgrade guide (for the whole v14 release)
- all models have their interfaces extracted to `Kentico.Kontent.Delivery.Abstractions` and we only return interfaces
  - for Asset we have IAsset, for ContentType there is an IContentType, etc.
- `DeliveryCacheManager` was renamed to `MemoryCacheManager`
- there is a new implementation of `IDeliveryCacheManager` called `DistributedCacheManager` which implements the `IDistributedCache` interface using [BSON serialization](http://bsonspec.org/)
  - it's possible to register the cache using `Kentico.Kontent.Delivery.Caching.Extensions.ServiceCollectionExtensions.AddDeliveryClientCache()` by changing `DeliveryCacheOptions.CacheType` from `Memory` to `Distributed`. by default, it registers the `MemoryDistributedCache`. if you want to use a different implementation (e.g. redis, you need to register its instance before calling `AddDeliveryClientCache()`
- `IPropertyValueConverter.GetPropertyValue` now is now strongly typed. Instead of `JToken`, you receive `ContentElementValue<T>` with Name, Codename, Type, and Value properties where Value is of type T and T is the type of your property (DateTime, string, int, Asset....). also, the method is async now.
- methods in `IContentLinkUrlResolver`, `IModelProvider`, `IInlineContentItemsProcessor` are now async, their signatures remain the same
- some models are now more specific (contain e.g. `Guid` instead of `string` where it was appropriate)
- places which returned `ContentElement` now return `IContentElement`. plus, based on the type of the element, they can return a type castable to `IMultipleChoiceElement` or `ITaxonomyElement` to allow strongly typed access to members specific to these types
- the `IDeliveryClient` now contains only async methods that operate upon strongly-typed models. all JSON-based methods were removed. if someone wishes to access the raw json, all `Delivery*Response` objects contain an object called ApiResponse of the `IApiResponse` type. this property contains low-level data like `string Content`, `string RequestUrl`, or `string ContinuationToken`.

### Notes for the reviewer
- we got rid of the static `JsonConvert` and we inject a custom serializer instance via DI (see `services.TryAddSingleton<JsonSerializer>(new DeliveryJsonSerializer());` in `ServiceCollectionExtensions`
- this custom instance - `DeliveryJsonSerializer` - facilitates the instantiation of concrete types for interface type models via `DeliveryContractResolver` and `DeliveryServiceCollection`. it also facilitates translation to the best-suited models for content elements (see the `ContentElementConverter`)
- all concrete model types have been marked as internal
  - their deserialization was unified. we now only use `.ToObject<T>()` and don't use anything `JToken` or `JObject`-related. a lot of code was relying on JTokens, now we use a strongly-typed approach. also, all model constructors were unified to allow this kind of deserialization.
- one of such places was `RichTextContentConverter` and all kinds of `IPropertyValueConverter`s
- getting rid of Newtonsoft.Json types from models and interfaces allowed to free the `Kentico.Kontent.Delivery.Abstractions` from all non-framework dependencies. also, note that all types reside in the main namespace. this makes it easy to use them anywhere without injecting multiple using statements.
- `IReadOnlyList` occurrences were changed to `IList` to allow (de)serialization of the whole Delivery*Response types (the same goes with  `IReadOnlyDictionary` and `IDictionary`)
- distributed caching implementation details:
  - the BSON serialization loses precision for DateTime formats, therefore unit tests are decorated with `AssertionExtensions.DateTimesBsonCorrection` which allows 1ms difference
  - also, the implementation of `IDistributedCache` is much simpler than the other one. so it doesn't allow invalidating by cache dependencies, only by cache keys. this is reflected in the unit tests using `[InlineData(CacheTypeEnum.Memory)]` and `[InlineData(CacheTypeEnum.Distributed)]` for the given theories
  - for tests, we use the `MemoryDistributedCache`
- another type that is now being injected via DI is the `IHtmlParser`. creating an instance of this type is quite costly, so it should save some resources
- the `ContentLinkResolver` has been refactored to support async
- the Delivery*Response model types have been freed of all logic
- the `InlineContentItemsProcessor` now relies on `ParseDocumentAsync` instead of its sync `Parse` counterpart
- the `Content` property in the `ApiResponse` allows access to low-level data. it might make sense to switch it to an async method. any thoughts?

#### Code benchmarks
- there is a new project called `Kentico.Kontent.Delivery.Benchmarks`. you can run the benchmarks by switching to the "Release" build configuration and running the respective unit tests or by compiling in the Release mode and running the resulting exe. the results of the benchmarks are in the 'Kentico.Kontent.Delivery.Benchmarks\bin\Release\netcoreapp3.1\BenchmarkDotNet.Artifacts' folder

**Results:**
v13.0.1:
Benchmark Net 13 0 1

v14.0.0-beta12:
Benchmark Net 14 0 0-beta12

There is a notable performance drop when it comes to the SDK code-only benchmark. My suspicion is that there is some overhead related to `async`. We might want to profile and investigate the exact causes further.

#### Load tests
Configuration: Release
Host: Kestrel
Tool: [Bombardier](https://github.com/codesenberg/bombardier)
Site: [Kentico Kontent .NET Boilerplate](https://github.com/Kentico/kontent-boilerplate-net)
LogLevel: Warning
Caching: off

PS1: `.\bombardier.exe -c 128  -d 10s -l https://localhost:5001`
PS2: `dotnet counters monitor --process-id <PID> System.Runtime Microsoft.AspNetCore.Hosting`

v13.0.1:
10s_13

v14.0.0-beta12:
10s_14

The throughput and latency distribution of both versions are comparable and within the margin of error. So this good news.

#### Conclusions & suggestions
My suggestion is to release an RC version of the SDK and let users test it in real-world scenarios. In the meantime, we can profile the code and run more sophisticated, composite benchmarks in isolated environments and keep improving the code.

### What to verify
- test together with the new generator: https://github.com/Kentico/kontent-generators-net/releases/tag/6.0.0-beta4
- try migrating an existing app (own or https://github.com/Kentico/kontent-sample-app-razorpages or https://github.com/Kentico/kontent-boilerplate-net)
- test against the live API
- verify the list of breaking changes 
- verify the perf tests (see the previous chapter)
- verify that the developer experience hasn't been impacted negatively


tomasjurasek:
This bunch of changes is very good - async API, more extensible, and modular. 

I agree with you about the release of some RC version for testing purposes, but I'm a little scared about doing a perf testing in this big bunch of changes.

petrsvihlik:
### Benchmarks against the production server (not against test fixtures)

**v13.0.1**
|     Method |     Mean |    Error |   StdDev | Gen 0 | Gen 1 | Gen 2 |  Allocated |
|-------------- |---------:|---------:|---------:|------:|------:|------:|-----------:|
|  GetItemAsync | 34.72 ms | 1.237 ms | 3.609 ms |     - |     - |     - |  248.43 KB |
| GetItemsAsync | 39.96 ms | 1.119 ms | 3.247 ms |     - |     - |     - | 1467.77 KB |


**v14.0.0-beta12**
|        Method |     Mean |    Error |   StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|-------------- |---------:|---------:|---------:|------:|------:|------:|----------:|
|  GetItemAsync | 36.83 ms | 1.025 ms | 2.941 ms |     - |     - |     - |   1.46 MB |
| GetItemsAsync | 43.16 ms | 1.208 ms | 3.327 ms |     - |     - |     - |   4.47 MB |

So the difference here is only in the memory allocated. We'll continue the investigation.

### Snapshots from .NET Counters from bombarding the site for many minutes:

**v13.0.1**
13_counters2
- memory allocation was usually around 700MB (1,5GB max)

**v14.0.0-beta12**
14_counters
- memory allocation was usually around 400MB

PR #254, opened 2021-01-08
Feature/service collection extensions

### LINKED ISSUE #240: Spike: An alternative ServiceCollectionExtensions for named services
### Motivation

Although it's convenient for the consumers and tempting for authors to offer support for registering named services out of the box, e.g.:
- `public static IServiceCollection AddDeliveryClient(this IServiceCollection services, string name, IConfiguration configuration, string configurationSectionName = "DeliveryOptions")`
- `public static IServiceCollection AddDeliveryClientCache(this IServiceCollection services, string name, DeliveryCacheOptions options)`

it shouldn't really be the responsibility of this SDK to facilitate complex DI resolution scenarios.

**Current problems that we see:**
- not to introduce another NuGet dependency, we are solving the problem with [factories](https://stackoverflow.com/a/44177920/1332034) which is a workaround for the ASP.NET Core's default DI container
- it doesn't solve the problem of nested dependencies completely (scenarios such as using different type providers, link and rich-text resolvers, etc.)
- lots of code to maintain and scenarios to be covered with unit tests

### Proposed solution

- Create an extra NuGet for complex scenarios that would contain more advanced `ServiceCollectionExtensions`.
- Use `InternalsVisibleTo` to enable access to internal classes from `Kentico.Kontent.Delivery` (it's a rare use case for that attribute but I think justifiable in this case)
  - http://www.jasonbock.net/JB/News/Item/4b3f9157-c247-4de5-8850-4d91c79ef172
  - https://stackoverflow.com/questions/3279639/when-should-assembly-internalsvisibleto-be-used
- Make it dependent on Autofac and use [named or keyed services](https://autofaccn.readthedocs.io/en/latest/advanced/keyed-services.html)
- Follow the solution outlined [here](https://sethflowers.com/2017/03/22/AutoFac-dynamic-resolution-of-service-multiple-levels-deep.html) and [here](https://stackoverflow.com/questions/6262704/how-to-resolve-interface-based-on-service-where-its-passed-to):

```c#
public class TestGrandChildResolutions
{
    public interface IA1 { IB B { get; set; } }
    public class A1 : IA1 { public IB B { get; set; } public A1(IB b) { this.B = b; } }

    public interface IA2 { IB B { get; set; } }
    public class A2 : IA2 { public IB B { get; set; } public A2(IB b) { this.B = b; } }

    public interface IB { IC C { get; set; } }
    public class B : IB { public IC C { get; set; } public B(IC c) { this.C = c; } }

    public interface IC { }
    public class C1 : IC { }
    public class C2 : IC { }

    [Test]
    public void Test()
    {
        ContainerBuilder builder = new ContainerBuilder();

        builder.RegisterType<C1>().AsImplementedInterfaces().Named<IC>("C1-Name");
        builder.RegisterType<C2>().AsImplementedInterfaces().Named<IC>("C2-Name");

        builder.RegisterType<B>().AsImplementedInterfaces().WithParameter(ResolvedParameter.ForNamed<IC>("C1-Name")).Named<IB>("B1-Name");
        builder.RegisterType<B>().AsImplementedInterfaces().WithParameter(ResolvedParameter.ForNamed<IC>("C2-Name")).Named<IB>("B2-Name");

        builder.RegisterType<A1>().AsImplementedInterfaces().WithParameter(ResolvedParameter.ForNamed<IB>("B1-Name"));
        builder.RegisterType<A2>().AsImplementedInterfaces().WithParameter(ResolvedParameter.ForNamed<IB>("B2-Name"));

        IContainer container = builder.Build();
        IA1 a1 = container.Resolve<IA1>();
        IA2 a2 = container.Resolve<IA2>();

        Assert.IsInstanceOf<C1>(a1.B.C, "C1 Check");
        Assert.IsInstanceOf<C2>(a2.B.C, "C2 Check");
    }
}
```
- and finally, remove the existing implementation of named services

### Additional context
- We should then document the new library and offer best practices.
- Once we validate that this approach works, we should get rid of the current named services support in the SDK.
- [A comment](https://github.com/Kentico/kontent-delivery-sdk-net/issues/239#issuecomment-732431968) by @vletierce 
> I have been using the SDK over a year now and must say the fact that I can't assigned a custom type provider to a named HttpClient through the .NetCore dependency injection mechanisme represents a real problem on my side.
> There is the possibility to create an instance of a SharedTypeProvider class however it only works if all your content types (accross projects) have different names, unfortunately in my case I have content types that share the same name accross projects.
> 
> Could we have quick exemple of what solution 4 would look like ? In my case it is very important that I can match a content type provider to a specific named client.
> 
> Thanks

tomasjurasek:
### Motivation

Which issue does this fix? Fixes #240 

### Checklist

- [x] Tests are passing

### How to test


petrsvihlik (changes requested):
I'd remove the existing implementation and replace it with an Autofac-based one (as an additional package).

tomasjurasek:
> I'd remove the existing implementation and replace it with an Autofac-based one (as an additional package).

The original implementation is removed now. I also removed the 'InternalsVisibleTo' parameter for the new package and create some factories for returns these internal implementations.  I need to do some tests for these new use cases.

tomasjurasek:
> > I'd remove the existing implementation and replace it with an Autofac-based one (as an additional package).
> 
> The original implementation is removed now. I also removed the 'InternalsVisibleTo' parameter for the new package and create some factories for returns these internal implementations. I need to do some tests for these new use cases.

The new package does not depend on internal references. The infrastructure is open for new packages around DI.

Kentico.Kontent.Delivery.Extensions.Autofac.DependencyInjection/DeliveryClientFactory.cs
```diff
@@ -0,0 +1,88 @@
...
+            {
+                throw new ArgumentNullException(nameof(name));
+            }
+
+            if (!_cache.TryGetValue(name, out var client))
+            {
+                var deliveryClientOptions = _deliveryOptions.Get(name);
+
+                // Validate that the option object is indeed configured
+                if (deliveryClientOptions.ProjectId != null)
```
petrsvihlik, line 47: This check should be also obsolete with the transition to Autofac

Kentico.Kontent.Delivery.Extensions.Autofac.DependencyInjection/DeliveryClientFactory.cs
```diff
@@ -0,0 +1,88 @@
...
+using Kentico.Kontent.Delivery.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+using Newtonsoft.Json;
+using System;
+using System.Collections.Concurrent;
+
+namespace Kentico.Kontent.Delivery.Extensions.Autofac.DependencyInjection
+{
+    internal class DeliveryClientFactory : IDeliveryClientFactory
```
petrsvihlik, line 13: for the sake of simplicity, I'd rename this implementation of IDeliveryClientFactory to something that cannot be mistaken with the default implementation...so what about `NamedDeliveryClientFactory`

Kentico.Kontent.Delivery/Extensions/DeliveryOptionsExtensions.cs
```diff
@@ -2,8 +2,16 @@
 
 namespace Kentico.Kontent.Delivery.Extensions
 {
-    internal static class DeliveryOptionsExtensions
+    /// <summary>
+    /// A class which contains extension methods on <see cref="DeliveryOptions"/>.
+    /// </summary>
+    public static class DeliveryOptionsExtensions
```
petrsvihlik, line 8: shouldn't we just move this one to the abstractions lib?

Kentico.Kontent.Delivery/Helpers/DeliveryOptionsHelpers.cs
```diff
@@ -0,0 +1,23 @@
...
+    /// A class which contains helper methods on <see cref="DeliveryOptions"/>.
+    /// </summary>
+    public static class DeliveryOptionsHelpers
+    {
+        /// <summary>
+        /// Builds the <see cref="DeliveryOptions"/> instance from the delegate.
+        /// </summary>
+        /// <param name="buildDeliveryOptions"></param>
+        /// <returns></returns>
+        public static DeliveryOptions Build(Func<IDeliveryOptionsBuilder, DeliveryOptions> buildDeliveryOptions)
```
petrsvihlik, line 17: Can't we just somehow abstract this and retrieve the `IDeliveryOptionsBuilder`?

Kentico.Kontent.Delivery.Extensions.Autofac.DependencyInjection.Tests/ServiceCollectionExtensionsTests.cs
```diff
@@ -0,0 +1,126 @@
...
+using Kentico.Kontent.Delivery.Extensions.Autofac.DependencyInjection.Extensions;
+using System;
+using Xunit;
+using FluentAssertions;
+using Autofac;
+using FakeItEasy;
+
+namespace Kentico.Kontent.Delivery.Extensions.Autofac.DependencyInjection.Tests
+{
+    public class ServiceCollectionExtensionsTests
```
petrsvihlik, line 13: can we add a test testing the multi-`ITypeProvider` scenario?

Kentico.Kontent.Delivery.Extensions.Autofac.DependencyInjection/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,127 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+using System;
+
+namespace Kentico.Kontent.Delivery.Extensions.Autofac.DependencyInjection.Extensions
+{
+    /// <summary>
+    /// A class which contains extension methods on <see cref="IServiceCollection"/> for registering an <see cref="IDeliveryClient"/> instance.
+    /// </summary>
+    public static class ServiceCollectionExtensions
```
petrsvihlik, line 18: can we somehow make it easier to register the required named services such as `ITypeProvider`, and the others like inlince content resolvers? Perhaps one method that would accept all the services and whose parameters would be null by default and the only required param would be `name` and that would register all the services under the given name?

petrsvihlik (changes requested):

Kentico.Kontent.Delivery.Extensions.Autofac.DependencyInjection/DeliveryClientFactory.cs
```diff
@@ -0,0 +1,88 @@
...
+
+            return client;
+        }
+
+        public IDeliveryClient Get() => GetService<IDeliveryClient>();
+
+        private IDeliveryClient Build(DeliveryOptions options, string name)
+        {
+            return Delivery.DeliveryClientFactory.Create(
+                new DeliveryOptionsMonitor(options, name),
```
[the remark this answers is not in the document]
tomasjurasek: > I think we might be able to get rid of the `DeliveryOptionsMonitor` and make it work similarly to the other services...
> `GetNamedServiceOrDefault<IOptions<DeliveryOptions>>(name)`

The GetNamedServiceOrDefault returns the named service of default value if is registered. I think this scenario is not acceptable with the Options setting. Do you want to do the same as with services?

Kentico.Kontent.Delivery/Helpers/DeliveryOptionsHelpers.cs
tomasjurasek: This is just a helper that evaluates `IDeliveryOptionsBuilder`.

Kentico.Kontent.Delivery.Extensions.Autofac.DependencyInjection.Tests/ServiceCollectionExtensionsTests.cs
tomasjurasek: I do.

Kentico.Kontent.Delivery/Extensions/DeliveryOptionsExtensions.cs
tomasjurasek: Yes, I do.

Kentico.Kontent.Delivery.Extensions.Autofac.DependencyInjection/DeliveryClientFactory.cs
tomasjurasek: Removed

tomasjurasek: The Autofac has no effects on how `_deliveryOptions `works

Kentico.Kontent.Delivery.Extensions.Autofac.DependencyInjection/DeliveryClientFactory.cs
tomasjurasek: Renamed

Kentico.Kontent.Delivery.Extensions.Autofac.DependencyInjection/Extensions/ServiceCollectionExtensions.cs
petrsvihlik: after an internal discussion: we'll leave that up to the consumer. we just need to document it well.
if needed, we can add something later.

Kentico.Kontent.Delivery.Extensions.Autofac.DependencyInjection/DeliveryClientFactory.cs
petrsvihlik: I remember this was a workaround that we introduced at about the time we implemented the named clients. The problem is that `Get(name)` never returns null. If there is no object of the given name, it returns a default instance.

Kentico.Kontent.Delivery.Extensions.Autofac.DependencyInjection/DeliveryClientFactory.cs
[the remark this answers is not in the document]
petrsvihlik: hmmm..probably not...i guess :D

Kentico.Kontent.Delivery.Extensions.Autofac.DependencyInjection/DeliveryClientFactory.cs
tomasjurasek: I changed this behavior. The option is bounded to its name and the name is compared with the name property. It does not depend on some option property value.

petrsvihlik: I hope the `internal` property won't cause any problems while (de)serializing the `DeliveryOptions` in ASP.NET apps or in the code generator.

tomasjurasek: I will do some testing with the real app.

Kentico.Kontent.Delivery.Extensions.DependencyInjection/ICustomServiceProvider.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace Kentico.Kontent.Delivery.Extensions.DependencyInjection
+{
+    /// <summary>
+    /// A class for providing a custom dependency.
+    /// </summary>
+    public interface ICustomServiceProvider
```
petrsvihlik, line 6: I'd be a bit more specific and called this INamedServiceProvider

petrsvihlik (changes requested):

petrsvihlik:
It looks good to me. Besides the one comment above, I think it's ready to be released as beta and tested.

PR #281, opened 2021-09-23
Return codename strings instead of "null" when object detail not retu…

### LINKED ISSUE #279: Return empty objects with IDs where "depth" parameter prevents object return
### Motivation

Returning a content item, which has a linked item element, with a depth of 1, will return a typed object which appears to have no items in that property, even though the underlying API call has returned the IDs (but not the content) of the linked items. There isn't currently a way to access those IDs to make the additional API calls to retrieve them.

### Proposed solution

We could return a "ghost" object which has no content except the ID for items which are linked but whose detailed content is missing from the API call return.

rossb-mmt:
Return codename strings instead of "null" when object detail not returned in the modular_content

### Motivation

Fixes #279 

Returns the codename string instead of null (or a messy skeleton object that looks like a content item with no content) for items that are excluded from the modular_content by the depth parameter on the API call. This means that the website can use that codename to later retrieve the detail for the content item if it is needed.

### Checklist

- [X] Code follows coding conventions held in this repo
- [X] Automated tests have been added
- [X] Tests are passing
- [X] Temporary settings (e.g. variables used during development and testing) have been reverted to defaults

### How to test

### DIFF
--- a/Kentico.Kontent.Delivery.Tests/DeliveryClientTests.cs
+++ b/Kentico.Kontent.Delivery.Tests/DeliveryClientTests.cs
@@ -105,7 +105,48 @@ namespace Kentico.Kontent.Delivery.Tests
             Assert.NotEmpty(beveragesResponse.Item.TeaserImage);
             Assert.NotEmpty(beveragesResponse.Item.Personas);
         }
+        [Fact]
+        public async Task GetLinkedItemsWithDepthZeroReturnsCodenames()
+        {
+            // Arrange
+            _mockHttp
+                .When($"{_baseUrl}/items/home")
+                .Respond("application/json",
+                    await File.ReadAllTextAsync(Path.Combine(Environment.CurrentDirectory,
+                        $"Fixtures{Path.DirectorySeparatorChar}home_depth0.json")));
+            var client = InitializeDeliveryClientWithCustomModelProvider(_mockHttp, new PropertyMapper(), new CustomTypeProvider());
 
+            // Act
+            var homeResponse = await client.GetItemAsync<Home>("home");
+
+            // Assert
+            Assert.True(homeResponse?.ApiResponse?.IsSuccess);
+            Assert.NotEmpty(homeResponse.Item.HeroUnit);
+            Assert.Equal(2, homeResponse.Item.HeroUnit.Count());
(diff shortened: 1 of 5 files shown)


MiroKentico:
Hi rossb-mmt, I will check this next Monday or Tuesday.

MiroKentico:
Hi @rossb-mmt, I looked into your changes, @Simply007 also took a brief look into this. We need deeper solution analysis. I will keep you posted.

rossb-mmt:
> Hi @rossb-mmt, I looked into your changes, @Simply007 also took a brief look into this. We need deeper solution analysis. I will keep you posted.

Thanks. I'm happy to help out with a more elaborate solution if you need it.

MiroKentico:
> 
> 
> > Hi @rossb-mmt, I looked into your changes, @Simply007 also took a brief look into this. We need deeper solution analysis. I will keep you posted.
> 
> Thanks. I'm happy to help out with a more elaborate solution if you need it.

Hi @rossb-mmt, thanks for your offer, I will keep it in mind. Right now, there are few priority topics (like fix of some flaky tests), but we definitely plan to solve your issue.

MiroKentico:
We discussed solution proposed in this PR with @Simply007. To be honest, we would prefer to avoid change of type of linked item model (object) to string. This is in fact breaking change. Therefore we would like to explore additional possible solutions, for instance we might want to check whether

- usage https://github.com/Kentico/kontent-delivery-sdk-net/wiki/Support-for-custom-types-in-models-via-Value-Converters could cover your use case
- there is a reasonable way to put linked items codenames into addtional custom property with type IEnumerable<string>

rossb-mmt:
> We discussed solution proposed in this PR with @Simply007. To be honest, we would prefer to avoid change of type of linked item model (`object`) to `string`. This is in fact breaking change. Therefore we would like to explore additional possible solutions, for instance we might want to check whether
> 
> * usage https://github.com/Kentico/kontent-delivery-sdk-net/wiki/Support-for-custom-types-in-models-via-Value-Converters could cover your use case
> * there is a reasonable way to put linked items codenames into addtional custom property with type `IEnumerable<string>`

Thanks, I completely understand what you say about it being a breaking change.

Right now the workaround is just to increase the Depth parameter on the call, which is wasteful.

PR #299, opened 2021-11-11
284 adjusted namespaces to be like folder structure

MiroKentico:
Adjust namespaces (and usings) to reflect folder structure. Some of them were already adjusted in base branch 248_dependency_issues. 

I did it via Rider IDE automatic functionality and randomly verified on few files modified.


Sevitas (approved):
Number of tests is equal
IDE doesn't recommend any namespace changes
Solution is buildable
briefly reviewed files

Issue #284, opened 2021-10-14
ImageTransformation - dependency issues


petrsvihlik:
### Brief bug description
getting these errors
> error CS0433: The type 'ImageUrlBuilder' exists in both 'Kentico.Kontent.ImageTransformation, Version=14.2.1.0, Culture=neutral, PublicKeyToken=null' and 'Kentico.Kontent.Urls, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null'

### Repro steps

- reference both the latest version of the:
  - Kentico.Kontent.AspNetCore
  - Kentico.Kontent.Delivery
- try to use ImageUrlBuilder or ImageFormat

### Stuff to fix
- re-release https://github.com/Kentico/kontent-aspnetcore with the latest dependency on Kentico.Kontent.Delivery
- https://github.com/Kentico/kontent-delivery-sdk-net/tree/master/Kentico.Kontent.ImageTransformation -> rename the folder to match the project file

Simply007:
Steps to fix are OK - thx @petrsvihlik:
* Rename `~/Kentico.Kontent.ImageTransformation` folder to `~/Kentico.Kontent.Urls` and release new minor version
* Upgrade https://github.com/Kentico/kontent-aspnetcore with newly released version `0.12.0` - you can release beta first to test the output
** maybe @petrsvihlik can test the beta out hm? 😉

petrsvihlik:
I guess this one is blocked by #292 right?

Simply007:
> I guess this one is blocked by #292 right?

We have found a workaround for the releases (#294). But we wanted to walk through all namespaces and fix potential problems and include them into the new major version - #287.

Co currently it is being blocked by #287.

petrsvihlik:
right...but I guess that should be pretty easy, right? I think even VS can adjust all namespaces in the whole solution with a single click and then you can just check the git changes to verify. or is there something more tricky?

Simply007:
> right...but I guess that should be pretty easy, right? I think even VS can adjust all namespaces in the whole solution with a single click and then you can just check the git changes to verify. or is there something more tricky?

Yeap, it was just not at the top of our backlog. Do you have any estimate, you would like to have it fixed?

petrsvihlik:
naah..I was just wondering. I might draft the PR myself if it'd help ya

Simply007:
> naah..I was just wondering. I might draft the PR myself if it'd help ya

[...]

PR #300, opened 2021-11-12
Namespace unifications

### LINKED ISSUE #287: Unify namespaces
### Motivation

Unify namespaces:
- some entities have namespace matching their full project structure
- some entities match just project's namespace

### Proposed solution

Unify namespaces throughout the whole solution structure/match the project namespace in all classes

Simply007:
### Motivation

Fixes #287 

### Checklist

- [ ] Tests are passing

### How to test


petrsvihlik:
please don't do this...leaving abstractions in the root namespace was a conscious and intentional decision to make the whole API easily discoverable. also, you'll create a huge breaking change again + you'd need to update the code generator, etc.

https://github.com/Kentico/kontent-delivery-sdk-net/releases/tag/14.0.0


I'm already facing a problem because of this because I have dependencies targeting various versions of the library.
image

suggestion: revert this commit

Simply007:
> please don't do this...leaving abstractions in the root namespace was a conscious and intentional decision to make the whole API easily discoverable. also, you'll create a huge breaking change again + you'd need to update the code generator, etc.
> 
> https://github.com/Kentico/kontent-delivery-sdk-net/releases/tag/14.0.0
> 
> I'm already facing a problem because of this because I have dependencies targeting various versions of the library. image
> 
> suggestion: revert this commit

Fair point!

I tried to find any official ruleset for namespaces and folders, but unsuccessfully. In unofficial content, opinions vary. 

Could you elaborate more on your use case with multiple versions of the library that is causing the issues? 

According to the generator's topic, the adjustments are planned. For the end project, it should be one automatic fix to set proper namespaces out of generated code.

In [version 16](https://github.com/Kentico/kontent-delivery-sdk-net/milestone/20), there will be a breaking change with the target framework with at least one namespace changes for Image transformation (#284), so this is a good time to clean up the namespace and set the best practices for code.

I will discuss the pros and cons of this change with maintainers on Thursday (2021-11-18).

---

Saving the commit ID for future reference: https://github.com/Kentico/kontent-delivery-sdk-net/pull/300/commits/c5ea2044cb84e423ca331832030089a1cde21542

petrsvihlik:
>Could you elaborate more on your use case with multiple versions of the library that is causing the issues?

Yeah, it's caused by referencing both Kentico.Kontent.Delivery and Kontent.Statiq (which references K.K.Delivery [too](https://github.com/alanta/Kontent.Statiq/blob/main/Kontent.Statiq/Kontent.Statiq.csproj)) in my project. So any API breaking change between the two K.K.D. referenced is causing the errors.

>so this is a good time to clean up the namespace and set the best practices for code.

Yeah, that's what we said the last time when we decided to unify it this way. :)

Anyways, if you decide to go this path, I'd really suggest [making the generator part of this repo](https://github.com/Kentico/kontent-delivery-sdk-net/issues/249) and releasing it always together with fresh nugets.

Simply007:
> Yeah, that's what we said the last time when we decided to unify it this way. :)

We will stick with the original approach. I will revert the commit https://github.com/Kentico/kontent-delivery-sdk-net/commit/c5ea2044cb84e423ca331832030089a1cde21542 and add information about namespaces into README at least so that we won't do go this way again.

> Anyways, if you decide to go this path, I'd really suggest making the generator part of this repo and releasing it always together with fresh nugets.

Let's leave this decision for https://github.com/Kentico/kontent-delivery-sdk-net/issues/249

PR #303, opened 2021-11-21
Unify namespaces in Kentico.Kontent.Delivery.Abstractions

Simply007:
### Motivation

Unify all namespaces in `Kentico.Kontent.Delivery.Abstractions` assembly to the same namespace.

Based on https://github.com/Kentico/kontent-delivery-sdk-net/pull/300#issuecomment-974854350

Related to #287

### Checklist

- [x] Tests are passing

### How to test


Simply007:
@petrsvihlik - I have bad news - I have found one class -  `DeliveryOptionsExtensions` - that used sub-namespace `Kentico.Kontent.Delivery.Abstractions.Extensions`.

I wanted to fix [single namespace requirement and cover this rule in tests](https://github.com/Kentico/kontent-delivery-sdk-net/pull/303/files#diff-8ca36e753ec77fe4c007d98117f7ffdd6917d09590d281f7da18d3e5a806ca38).

For now, I have fixed this, but I would like to hear your opinion on this first.

petrsvihlik:
Yeah, it makes sense to unify this one and have it covered by a unit test - a good idea btw.

Regarding the prevention of problems caused by referencing multiple versions of the same package (e.g. through nested dependencies). The only good solution is to [be 100% backwards compatible with older versions of the same package](https://azure.github.io/azure-sdk/dotnet_introduction.html#client-versions). This requires extreme diligence when designing the APIs and ideally a very thorough review process. 

Just for the sake of completeness - there is another way to prevent assembly conflicts and that's called [strong-naming](https://docs.microsoft.com/en-us/dotnet/standard/assembly/strong-named). I even wrote [an article](https://devnet.kentico.com/articles/referencing-multiple-versions-of-the-same-assembly-in-a-single-application) on that a couple of years ago. Even MSFT says that ["Strong-names are leftover from previous eras of .NET"](https://github.com/dotnet/runtime/blob/main/docs/project/strong-name-signing.md) and acknowledges that ["the majority of third-party assemblies do not need strong names"](https://docs.microsoft.com/en-us/dotnet/standard/assembly/create-use-strong-named). Even though, MSFT says that publicly published .NET libraries should be strong-named as not strong-naming a .NET library excludes anyone who does need to strong-name their application or library from using it. But even if you decide to start strong-naming the assemblies, you have no control over other publishers of libraries that depend on your library. You can't force them to reference [strong-named assembly](https://docs.microsoft.com/en-us/dotnet/standard/assembly/reference-strong-named).
In other words, it might be a good idea to start signing the assemblies in case someone wants to use them in an application they have 100% control of. But it's not a solution to this problem in general.

petrsvihlik (approved):

PR #319, opened 2022-02-10
279 - enable to write a second value converter for the same field

### LINKED ISSUE #279: Return empty objects with IDs where "depth" parameter prevents object return

winklertomas:
### Motivation

Fixes #279. Enables to write a second value converter for the same field and adds a way to map to Kontent content item field without interfering with Newtonsoft JSON serializer. Can be used for example for linked item codenames as is demonstrated in this PR

### Checklist

- [ ] Tests are passing

### DIFF
--- a/Kentico.Kontent.Delivery/ContentItems/PropertyMapper.cs
+++ b/Kentico.Kontent.Delivery/ContentItems/PropertyMapper.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Linq;
 using System.Reflection;
 using Kentico.Kontent.Delivery.Abstractions;
 using Newtonsoft.Json;
@@ -24,6 +25,13 @@ namespace Kentico.Kontent.Delivery.ContentItems
             {
                 // If JsonIgnore is set, do not match
                 return false;
+            }
+
+            PropertyNameAttribute fieldNameAttr = modelProperty.GetCustomAttribute<PropertyNameAttribute>();
+            if (fieldNameAttr != null)
+            {
+                // Try to get the name of the field from the field name attribute
+                return fieldName.Equals(fieldNameAttr.PropertyName, StringComparison.Ordinal);
             }
 
             JsonPropertyAttribute propertyAttr = modelProperty.GetCustomAttribute<JsonPropertyAttribute>();
--- a/Kentico.Kontent.Delivery/ContentItems/PropertyMapper.cs
+++ b/Kentico.Kontent.Delivery/ContentItems/PropertyMapper.cs
@@ -27,13 +27,13 @@ namespace Kentico.Kontent.Delivery.ContentItems
                 return false;
             }
 
-            var propertyName = GetPropertyName(modelProperty);
+            var propertyName = GetPropertyNameFromAttribute(modelProperty);
             return propertyName != null
                 ? fieldName.Equals(propertyName, StringComparison.Ordinal)
                 : fieldName.Replace("_", "").Equals(modelProperty.Name, StringComparison.OrdinalIgnoreCase); // Default mapping
         }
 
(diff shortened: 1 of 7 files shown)


Simply007 (changes requested):
* rename `FieldNameAttribute` to `PropertyNameAttribute` (describe in XMLdoc)

Q: -> extendability of FieldName -> consequences, Current possibilities of JsonProperty
A: It is still possible to use JsonProperty

TODO review:
* distributed caching and binary serialization => check if we have tests for https://github.com/Kentico/kontent-delivery-sdk-net/wiki/Caching-responses#distributedcachemanager -> extend https://github.com/Kentico/kontent-delivery-sdk-net/blob/master/Kentico.Kontent.Delivery.Caching.Tests/TestItem.cs tie one property with custom attribute
* Describe the attribute: https://github.com/Kentico/kontent-delivery-sdk-net/wiki/Working-with-strongly-typed-models#naming-the-properties

winklertomas:
I've renamed the FieldNameAttribute to PropertyNameAttribute and added the attribute to the property of TestItem.cs to show that it doesn't affect caching... 

In my opinion, the wiki should be updated after the PR is merged to the master branch and it should look e.g. as follows:

If you need to change the codename of an element that a property is bound to, you can enrich the property with the Kentico.Kontent.Delivery.ContentItems.PropertyNameAttribute attribute.

[PropertyName("text_field")]
public string ArticleTitle { get; set; }

If you also want to set the behavior of the Newtonsoft and don't want to map the same element to different properties, you can use the Newtonsoft.Json.JsonPropertyAttribute.

[JsonProperty("text_field")]
public string ArticleTitle { get; set; }

Kentico.Kontent.Delivery/ContentItems/PropertyNameAttribute.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+
+namespace Kentico.Kontent.Delivery.ContentItems
```
MiroKentico, line 3: What about move of this whole class to namespace (and appropriate folder) `Kentico.Kontent.Delivery.ContentItems.Attributes` ? The benefit would be better structure.

I am thinking about this, because for instance some custom attributes for RichText are already in separate namespace `Kentico.Kontent.Delivery.ContentItems.RichText.Attributes` (and not in root of `...RichText`).

MiroKentico (changes requested):
There are few comments, please check them.

Kentico.Kontent.Delivery/ContentItems/PropertyNameAttribute.cs
MiroKentico: Thanks for changing this.

Kentico.Kontent.Delivery/ContentItems/PropertyMapper.cs
```diff
@@ -26,6 +27,13 @@ public bool IsMatch(PropertyInfo modelProperty, string fieldName, string content
                 return false;
             }
 
+            PropertyNameAttribute fieldNameAttr = modelProperty.GetCustomAttribute<PropertyNameAttribute>();
```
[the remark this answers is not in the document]
MiroKentico: Thanks @Sevitas for implementation,

MiroKentico (approved):
Comments were applied, approve.

PR #328, opened 2022-05-12
Proper DeliveryAPI api error content is now forwarded instead of failing with null exception.

### LINKED ISSUE #326: GetItemsAsync and GetTaxonomiesAsync return "Object reference not set to an instance of an object" rather than a helpful error message, when the Project ID or ApiKey is incorrect
### Brief bug description

GetItemsAsync and GetTaxonomiesAsync both return {"Object reference not set to an instance of an object"} error message when the project ID or ApiKey is incorrect. This makes error handling much more difficult as it does not explain the cause.

### Repro steps

1. Create a delivery client with an invalid project id (e.g. if the project has just been re-cloned but we have forgotten to update the settings in our code)

2. Call GetTaxonomiesAsync or GetItemsAsync

e.g.    response = await deliveryClient.GetTaxonomiesAsync(new EqualsFilter("system.codename", taxonomyGroupName));

### Expected behavior

A more helpful error message, like the one below returned from GetItemAsync in the same situation:

"APIResponse returned error code: '3' APIResponse returned message: 'Missing or invalid access token. Please include the valid access token value in the Authorization header field as an HTTP bearer authorization scheme.'"

### Test environment

 - Platform: .NET Core 3.1

### Additional context

**GetItemsAsync error message and stack trace:**
  
  {"Object reference not set to an instance of an object."}   System.NullReferenceException

   Source: "Kentico.Kontent.Delivery"
   
```
   at Kentico.Kontent.Delivery.DeliveryClient.<GetItemsAsync>d__11`1.MoveNext()
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at Services.Content.Kentico.ItemService.AgentPortal.AgentPortalCoreKenticoClient.<GetItemsAsync>d__2`1.MoveNext() in D:\Projects\Services.Content\Data\Services.Content.Kentico\ItemService\AgentPortal\AgentPortalCoreKenticoClient.cs:line 113
```

**GetTaxonomiesAsync error message and stack trace:**

  {"Object reference not set to an instance of an object."}   System.NullReferenceException
   
   Source: "Kentico.Kontent.Delivery"
   
   StackTrace: 
   
```
   at Kentico.Kontent.Delivery.DeliveryClient.<GetTaxonomiesAsync>d__17.MoveNext()
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at Services.Content.Kentico.ItemService.AgentPortal.AgentPortalCoreKenticoClient.<GetTaxonomyGroup>d__6.MoveNext() in D:\Projects\Services.Content\Data\Services.Content.Kentico\ItemService\AgentPortal\AgentPortalCoreKenticoClient.cs:line 220
```

MiroKentico:
### Motivation

Fixes #326

All `DeliveryClient` methods that reads data from Delivery API now forward proper API error content instead of failing with null exception.

This PR code additionally introduces few unit tests refactorings/enhancements + non breaking constructors changes.

### DIFF
--- a/Kentico.Kontent.Delivery/ContentItems/DeliveryItemResponse.cs
+++ b/Kentico.Kontent.Delivery/ContentItems/DeliveryItemResponse.cs
@@ -1,5 +1,4 @@
-﻿using System.Collections.Generic;
-using Kentico.Kontent.Delivery.Abstractions;
+﻿using Kentico.Kontent.Delivery.Abstractions;
 using Kentico.Kontent.Delivery.SharedModels;
 using Newtonsoft.Json;
 
--- a/Kentico.Kontent.Delivery/ContentItems/DeliveryItemResponse.cs
+++ b/Kentico.Kontent.Delivery/ContentItems/DeliveryItemResponse.cs
@@ -24,5 +24,13 @@ namespace Kentico.Kontent.Delivery.ContentItems
         {
             Item = item;
         }
+        
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DeliveryItemResponse{T}"/> class.
+        /// </summary>
+        /// <param name="response">The response from Kontent Delivery API that contains a content item.</param>
+        internal DeliveryItemResponse(ApiResponse response) : base(response)
+        {
+        }
     }
 }
--- a/Kentico.Kontent.Delivery.Tests/DeliveryClientTests.cs
+++ b/Kentico.Kontent.Delivery.Tests/DeliveryClientTests.cs
@@ -244,6 +244,25 @@ namespace Kentico.Kontent.Delivery.Tests
             Assert.True(response?.ApiResponse?.IsSuccess);
             Assert.NotEmpty(response.Items);
         }
+        
+        [Fact]
+        public async Task GetItemsAsync_InvalidProjectId_RespondsWithApiError()
+        {
+            var deliveryApiMockedResponse = await File.ReadAllTextAsync(Path.Combine(Environment.CurrentDirectory, $"Fixtures{Path.DirectorySeparatorChar}DeliveryClient{Path.DirectorySeparatorChar}apiError_invalid_project.json"));
+            deliveryApiMockedResponse = deliveryApiMockedResponse.Replace("{PROJECT_ID}", _guid.ToString());
+            
+            _mockHttp
+                .When($"{_baseUrl}/items")
+                .Respond(HttpStatusCode.NotFound, "application/json",deliveryApiMockedResponse);
+
+            var client = InitializeDeliveryClientWithACustomTypeProvider(_mockHttp);
+
+            var response = await client.GetItemsAsync<object>();
+
+            Assert.False(response?.ApiResponse?.IsSuccess);
+            Assert.NotNull(response?.ApiResponse?.Error);
+            //TODO maybe assertions should be more detailed (what error code, message ...)
+        }
 
         [Fact]
         public void GetItemsFeed_DepthParameter_ThrowsArgumentException()
--- a/Kentico.Kontent.Delivery.Tests/DeliveryClientTests.cs
+++ b/Kentico.Kontent.Delivery.Tests/DeliveryClientTests.cs
@@ -371,6 +371,24 @@ namespace Kentico.Kontent.Delivery.Tests
             Assert.Equal(6, items.Count);
             Assert.Equal(2, timesCalled);
         }
+        
+        [Fact]
+        public async Task GetItemsFeedAsync_InvalidProjectId_RespondsWithApiError()
+        {
+            var expectedError = CreateInvalidProjectIdApiError();
+            var response = CreateApiErrorResponse(expectedError);
+
+            _mockHttp
+                .When($"{_baseUrl}/items-feed")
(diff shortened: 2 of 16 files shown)


Sevitas (changes requested):
I have just few notes:
1. Have you considered creating another constructor for responses (eg. DeliveryItemListingResponse) with just an ApiResponse param? It could save us setting those null values in the client.
2. Why there is no _succes check_ for _GetItemAsync, GetTypeAsync, GetContentElementAsync, GetTaxonomyAsync_ methods?
3. I think it would be better to move _succes check_ before assignement to the content variable.

MiroKentico:
> I have just few notes:
> 
>     1. Have you considered creating another constructor for responses (eg. DeliveryItemListingResponse) with just an ApiResponse param? It could save us setting those null values in the client.
> 
>     2. Why there is no _succes check_ for _GetItemAsync, GetTypeAsync, GetContentElementAsync, GetTaxonomyAsync_ methods?
> 
>     3. I think it would be better to move _succes check_ before assignement to the content variable.

1. No, I haven't thought about these constructors for null values. I will think about it and reach you to discuss it.
2. The reason is, that implementation was already done in a way, that they already produced correct error response  ... and added unit tests prove it. I will think about putting that `IsSuccess` check, but if I will put it there, it would be more for some code standardization than pure necessity.
3. You are right, it would be better place (to avoid doing not necessary code in these error cases). I will move this check as close to actual response retrieval where the code place is sub-optimal (case that you wrote).

MiroKentico:
> > I have just few notes:
> > ```
> > 1. Have you considered creating another constructor for responses (eg. DeliveryItemListingResponse) with just an ApiResponse param? It could save us setting those null values in the client.
> > 
> > 2. Why there is no _succes check_ for _GetItemAsync, GetTypeAsync, GetContentElementAsync, GetTaxonomyAsync_ methods?
> > 
> > 3. I think it would be better to move _succes check_ before assignement to the content variable.
> > ```
> 
>     1. No, I haven't thought about these constructors for null values. I will think about it and reach you to discuss it.
> 
>     2. The reason is, that implementation was already done in a way, that they already produced correct error response  ... and added unit tests prove it. I will think about putting that `IsSuccess` check, but if I will put it there, it would be more for some code standardization than pure necessity.
> 
>     3. You are right, it would be better place (to avoid doing not necessary code in these error cases). I will move this check as close to actual response retrieval where the code place is sub-optimal (case that you wrote).



> I have just few notes:
> 
>     1. Have you considered creating another constructor for responses (eg. DeliveryItemListingResponse) with just an ApiResponse param? It could save us setting those null values in the client.
> 
>     2. Why there is no _succes check_ for _GetItemAsync, GetTypeAsync, GetContentElementAsync, GetTaxonomyAsync_ methods?
> 
>     3. I think it would be better to move _succes check_ before assignement to the content variable.



> > I have just few notes:
> > ```
> > 1. Have you considered creating another constructor for responses (eg. DeliveryItemListingResponse) with just an ApiResponse param? It could save us setting those null values in the client.
> > 
> > 2. Why there is no _succes check_ for _GetItemAsync, GetTypeAsync, GetContentElementAsync, GetTaxonomyAsync_ methods?
> > 
> > 3. I think it would be better to move _succes check_ before assignement to the content variable.
> > ```
> 
>     1. No, I haven't thought about these constructors for null values. I will think about it and reach you to discuss it.
> 
>     2. The reason is, that implementation was already done in a way, that they already produced correct error response  ... and added unit tests prove it. I will think about putting that `IsSuccess` check, but if I will put it there, it would be more for some code standardization than pure necessity.
> 
>     3. You are right, it would be better place (to avoid doing not necessary code in these error cases). I will move this check as close to actual response retrieval where the code place is sub-optimal (case that you wrote).

Points 1. and 3. are now implemented, for point 2. is opened discussion within SDK maintainer team.

Kentico.Kontent.Delivery/ContentItems/DeliveryItemResponse.cs
```diff
@@ -24,5 +24,13 @@ internal DeliveryItemResponse(ApiResponse response, T item) : base(response)
         {
```
Sevitas, line 24: using System.Collections.Generic; - unused

Sevitas (dismissed):
Just a nitpick, otherwise LGTM.

Kentico.Kontent.Delivery/ContentItems/DeliveryItemResponse.cs
MiroKentico: Removed.

PR #329, opened 2022-05-20
DeliveryEndpointUrlBuilder(DeliveryOptions deliveryOptions) marked obsolete + potential null fix

### LINKED ISSUE #327: NullReference in DeliveryEndpointUrlBuilder
While investigating an other issue, I came across this bug:

https://github.com/Kentico/kontent-delivery-sdk-net/blob/aa802518a22f25ef6f31c608964788f5a12b33af/Kentico.Kontent.Urls/Delivery/DeliveryEndpointUrlBuilder.cs#L33

The code incorrectly assumes `_deliveryOptionsMonitor` is initialized which is not the case if the object was constructed using the second constructor. 
I'd put up a PR for it but I'm not sure if the second constructor is worth having since it's not being referenced in the project at all.

MiroKentico:
### Motivation

Fixes #327

`DeliveryEndpointUrlBuilder(DeliveryOptions deliveryOptions)` marked as obsolete and fixed potential null situation when this constructor would be used and property CurrentDeliveryOptions would be use read.

### DIFF
--- a/Kentico.Kontent.Urls/Delivery/DeliveryEndpointUrlBuilder.cs
+++ b/Kentico.Kontent.Urls/Delivery/DeliveryEndpointUrlBuilder.cs
@@ -30,7 +30,7 @@ namespace Kentico.Kontent.Urls.Delivery
         {
             get
             {
-                return _deliveryOptionsMonitor.CurrentValue ?? deliveryOptions;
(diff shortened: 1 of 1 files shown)


MiroKentico:
It was wrong direction of merge ...

PR #331, opened 2022-05-26
 DeliveryEndpointUrlBuilder(DeliveryOptions deliveryOptions) marked obsolete + potential null fix #329 

### LINKED ISSUE #327: NullReference in DeliveryEndpointUrlBuilder

MiroKentico:
### Motivation

Fixes #327

`DeliveryEndpointUrlBuilder(DeliveryOptions deliveryOptions)` marked as obsolete and fixed potential null situation when this constructor would be used and property CurrentDeliveryOptions would be use read.

### DIFF
--- a/Kentico.Kontent.Urls.Tests/DeliveryEndpointUrlBuilderTests.cs
+++ b/Kentico.Kontent.Urls.Tests/DeliveryEndpointUrlBuilderTests.cs
@@ -1,0 +1,59 @@
+﻿using System;
+using Kentico.Kontent.Delivery.Abstractions;
+using Kentico.Kontent.Urls.Delivery;
+using Microsoft.Extensions.Options;
+using Xunit;
+
+namespace Kentico.Kontent.Urls.Tests;
+
+public class DeliveryEndpointUrlBuilderTests
+{
+    //as this is only place where mock would be used, just fake implementation class is created
+    //if more mocking is to be used, please use mocking library like NSubstitute instead
+    private class FakeOptionsMonitor<T> : IOptionsMonitor<T>
+    {
+        public T CurrentValue { get; }
+
+        public FakeOptionsMonitor(T options)
+        {
+            CurrentValue = options;
+        }
+        
+        public T Get(string name)
+        {
+            throw new NotImplementedException();
+        }
+
+        public IDisposable OnChange(Action<T, string> listener)
+        {
+            throw new NotImplementedException();
+        }
+    }
+
+    [Fact]
+    public void GetLanguagesUrl_ConstructedWithDeliveryOptions_GetsLanguagesUrl()
+    {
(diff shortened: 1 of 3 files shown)


Kentico.Kontent.Urls/Delivery/DeliveryEndpointUrlBuilder.cs
```diff
@@ -51,6 +51,7 @@ public DeliveryEndpointUrlBuilder(IOptionsMonitor<DeliveryOptions> deliveryOptio
         /// Initializes the URL builder using <see cref="IOptionsMonitor{TOptions}"/>. Ideal for simple scenarios where hot reloading of configuration is not required.
         /// </summary>
         /// <param name="deliveryOptions">The configuration object.</param>
+        [Obsolete]
```
Sevitas, line 54: I think it might be worth it to edit the documentation of the constructor as well

Sevitas (changes requested):
Just a few suggestion in the PR.

Kentico.Kontent.Urls/Delivery/DeliveryEndpointUrlBuilder.cs
MiroKentico: Okay, I will at enhance this attribute with description parameter (to use `IOptionsMontor` version instead)

Kentico.Kontent.Urls.Tests/DeliveryEndpointUrlBuilderTests.cs
```diff
@@ -0,0 +1,57 @@
+﻿using System;
+using Kentico.Kontent.Delivery.Abstractions;
+using Kentico.Kontent.Urls.Delivery;
+using Microsoft.Extensions.Options;
+using Xunit;
+
+namespace Kentico.Kontent.Urls.Tests;
+
+public class DeliveryEndpointUrlBuilderTests
```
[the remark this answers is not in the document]
MiroKentico: Well, when I checked code coverage, it seemed that coverage `DeliveryEndpointUrlBuilder` public methods were covered by different automatic tests.

Nevertheless, to properly idea of unit tests, I have added tests for other public methods. But only in variant where options are passed by IOptionsMontior constructor, not the obsolete one. Obsolete constructor is still tested only by `GetItemUrl_ConstructedWithDeliveryOptions_ReturnsItemUrl()` test method.

PR #332, opened 2022-05-26
Async keyword replaced by Task.FromResult creation of result object.

### LINKED ISSUE #330: Warning in AssetElementValueConverter
### Brief bug description

What went wrong?

`AssetElementValueConverter.GetPropertyValueAsync` is marked as async but does not use await.

### Repro steps

1. Clone the repository
2. Build the solution
see

```plain
C:\projects\kontent-delivery-sdk-net\Kentico.Kontent.Delivery\ContentItems\AssetElementValueConverter.cs(21,35): warning CS1998: This async method lacks 'await' operators and will run 
synchronously. Consider using the 'await' operator to await non-blocking API calls, or 'await Task.Run(...)' to do CPU-bound work on a background thread. [C:\projects\kontent-delivery-sdk-net\Kentico.Kontent.Delivery\Kentico.Kontent.Delivery.csproj]
C:\projects\kontent-delivery-sdk-net\Kentico.Kontent.Delivery\ContentItems\AssetElementValueConverter.cs(21,35): warning CS1998: This async method lacks 'await' operators and will run 
synchronously. Consider using the 'await' operator to await non-blocking API calls, or 'await Task.Run(...)' to do CPU-bound work on a background thread. [C:\projects\kontent-delivery-sdk-net\Kentico.Kontent.Delivery\Kentico.Kontent.Delivery.csproj]
```

### Expected behavior

No warning

### Context

Related to #318

MiroKentico:
### Motivation

Fixes #330.

Solves issue from warning, that async code will be executed synchrnonoustly. As this code is in fact synchronous `async` keyword usage was replaced by `Task.FromResult` creation of result object (list of assets).

No new unit test was introduced as this should be covered by existing unit tests.

### DIFF
--- a/Kentico.Kontent.Delivery/ContentItems/AssetElementValueConverter.cs
+++ b/Kentico.Kontent.Delivery/ContentItems/AssetElementValueConverter.cs
@@ -18,7 +18,7 @@ namespace Kentico.Kontent.Delivery.ContentItems
             Options = options;
         }
 
-        public async Task<object> GetPropertyValueAsync<TElement>(PropertyInfo property, TElement contentElement, ResolvingContext context) where TElement : IContentElementValue<IEnumerable<IAsset>>
+        public Task<object> GetPropertyValueAsync<TElement>(PropertyInfo property, TElement contentElement, ResolvingContext context) where TElement : IContentElementValue<IEnumerable<IAsset>>
         {
             if (!typeof(IEnumerable<IAsset>).IsAssignableFrom(property.PropertyType))
             {
@@ -30,7 +30,7 @@ namespace Kentico.Kontent.Delivery.ContentItems
                 return null;
             }
 
-            return assetElementValue.Value
+            var assets = assetElementValue.Value
                 .Select(asset => new Asset
                 {
                     Description = asset.Description,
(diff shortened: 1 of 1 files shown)


Kentico.Kontent.Delivery/ContentItems/AssetElementValueConverter.cs
```diff
@@ -44,6 +44,8 @@ public AssetElementValueConverter(IOptionsMonitor<DeliveryOptions> options)
                 })
                 .Cast<IAsset>()
                 .ToList();
+
+            return Task.FromResult((object)assets);
```
Simply007, line 48: Is this a common practice @Sevitas @gormal @winklertomas?

gormal: If we need to return `Task` then it's ok. But do we need to return `Task`?

MiroKentico: Yes, as it is implemented now, we unfortunately do because this method is implementation of method from `public interface IPropertyValueConverter<in T>`.

To be honest, best solution would be be definitely to have lets say synchronous and asynchronous version of this method, but that would need more internal changes. We can of course discuss this.

Simply007 (approved):

PR #349, opened 2022-10-03
fix: update validation regex and related test API key

### LINKED ISSUE #348: Some API keys not passing the SDK validation
### Brief bug description

API keys in Kontent.ai are JWTs. As explained in [JWT docs](https://jwt.io/introduction), individual parts are encoded as base64url. [According to the specification](https://www.rfc-editor.org/rfc/rfc4648#section-5), base64url can contain dashes (-) and underlines (_). SDK [uses the following regex](https://github.com/kontent-ai/delivery-sdk-net/blob/master/Kontent.Ai.Delivery/Configuration/DeliveryOptionsValidator.cs#L12) to validate API keys: 

`new Regex(@"[A-Za-z0-9+/]+\.[A-Za-z0-9+/]+\.[A-Za-z0-9+/]+", RegexOptions.Compiled)`

The above regex will not pass for API keys, which contain either dash or underline right next to the dot which separates individual parts of the JWT, despite such API keys being valid. 

### Repro steps

1. Get an API key and adjust it according to the description
2. Pass the API key to UsePreviewApi() method during client instantiation
3. See error

### Expected behavior

The client should be instantiated properly.

### Test environment

Tested on version 16.0.0, which uses the same validation regex as the latest (17.0.0)

pokornyd:
### Motivation

Fixes #348

Test API key was updated manually, may not be a valid JWT but does illustrate the problem. Can provide a valid key that originally caused the issue in question.

### DIFF
--- a/Kontent.Ai.Delivery.Tests/Builders/Configuration/DeliveryOptionsBuilderTests.cs
+++ b/Kontent.Ai.Delivery.Tests/Builders/Configuration/DeliveryOptionsBuilderTests.cs
@@ -9,8 +9,8 @@ namespace Kontent.Ai.Delivery.Tests.Builders.Configuration
     {
         private const string ProjectId = "550cec62-90a6-4ab3-b3e4-3d0bb4c04f5c";
         private const string PreviewApiKey = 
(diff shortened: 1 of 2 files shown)


dzmitryk-kontent-ai (approved):
It seems that `+` and `/` characters in regex for base64url encoded jwt token are redundant, so I decided to get rid of them. Otherwise code review is ok. Merge it to master

PR #358, opened 2023-01-12
Make DistributedCacheManager more robust

### LINKED ISSUE #352: Make DistributedCacheManager more robust
### Brief bug description

DistributedCacheManager can't handle failures of the underlying cache implementation.

### Repro steps

1. Failing unit test: https://github.com/kontent-ai/delivery-sdk-net/blob/cacheClient_failing_tests/Kontent.Ai.Delivery.Caching.Tests/DeliveryClientCacheTests.cs#L681

### Expected behavior

DistributedCacheManager should fall back to the default client if the cache is unavailable.

`DeliveryCacheOptions` could be extended by one more property that would indicate that in case of failure of the underlying cache CacheManager would fall back to default `DeliveryClient`. Please note that this option should have an impact only on `DistributedCacheManager` implementation.

Also, consider logging as there should be a way to let customers know when their cache is unavailable.

dzmitryk-kontent-ai:
### Motivation

Which issue does this fix? Fixes #352

### Checklist

- [ ] Tests are passing

### How to test

### DIFF
--- a/Kontent.Ai.Delivery.Caching/DistributedCacheManager.cs
+++ b/Kontent.Ai.Delivery.Caching/DistributedCacheManager.cs
@@ -4,6 +4,7 @@ using System.Threading.Tasks;
 using Kontent.Ai.Delivery.Abstractions;
 using Kontent.Ai.Delivery.Caching.Extensions;
 using Microsoft.Extensions.Caching.Distributed;
+using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 
 namespace Kontent.Ai.Delivery.Caching
@@ -15,16 +16,21 @@ namespace Kontent.Ai.Delivery.Caching
     {
         private readonly IDistributedCache _distributedCache;
         private readonly DeliveryCacheOptions _cacheOptions;
+        private readonly ILogger _logger;
 
         /// <summary>
         /// Initializes a new instance of <see cref="DistributedCacheManager"/>
         /// </summary>
         /// <param name="distributedCache">An instance of an object that represent distributed cache</param>
         /// <param name="cacheOptions">The settings of the cache</param>
-        public DistributedCacheManager(IDistributedCache distributedCache, IOptions<DeliveryCacheOptions> cacheOptions)
+        /// <param name="loggerFactory">The factory used to create loggers.</param>
+        public DistributedCacheManager(IDistributedCache distributedCache,
+            IOptions<DeliveryCacheOptions> cacheOptions,
+            ILoggerFactory loggerFactory)
         {
             _distributedCache = distributedCache ?? throw new ArgumentNullException(nameof(distributedCache));
             _cacheOptions = cacheOptions.Value ?? new DeliveryCacheOptions();
+            _logger = loggerFactory.CreateLogger(nameof(DistributedCacheManager));
         }
 
         /// <inheritdoc />
@@ -40,6 +46,7 @@ namespace Kontent.Ai.Delivery.Caching
             }
             catch(Exception ex) when (ex is not ArgumentNullException && _cacheOptions.DistributedCacheResilientPolicy == DistributedCacheResilientPolicy.FallbackToApi) 
             {
+                _logger.LogInformation(ex, "Distributed cache is not available. Default DeliveryClient was used to get content from Delivery API.");
                 return await valueFactory();
             }
 
--- a/Kontent.Ai.Delivery.Caching/DistributedCacheManager.cs
+++ b/Kontent.Ai.Delivery.Caching/DistributedCacheManager.cs
@@ -5,6 +5,7 @@ using Kontent.Ai.Delivery.Abstractions;
 using Kontent.Ai.Delivery.Caching.Extensions;
 using Microsoft.Extensions.Caching.Distributed;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
 using Microsoft.Extensions.Options;
 
 namespace Kontent.Ai.Delivery.Caching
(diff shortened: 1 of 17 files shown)


Kontent.Ai.Delivery.Caching/DistributedCacheManager.cs
```diff
@@ -15,25 +16,38 @@ internal class DistributedCacheManager : IDeliveryCacheManager
...
         /// <summary>
         /// Initializes a new instance of <see cref="DistributedCacheManager"/>
         /// </summary>
         /// <param name="distributedCache">An instance of an object that represent distributed cache</param>
         /// <param name="cacheOptions">The settings of the cache</param>
-        public DistributedCacheManager(IDistributedCache distributedCache, IOptions<DeliveryCacheOptions> cacheOptions)
+        /// <param name="loggerFactory">The factory used to create loggers.</param>
+        public DistributedCacheManager(IDistributedCache distributedCache,
+            IOptions<DeliveryCacheOptions> cacheOptions,
+            ILoggerFactory loggerFactory)
```
gormal, line 29: Is it possible to make this optional? People could choose to opt-out then. Also, it would be more friendly from a backward compatibility point of view.

Kontent.Ai.Delivery.Caching/DistributedCacheManager.cs
```diff
@@ -15,25 +16,38 @@ internal class DistributedCacheManager : IDeliveryCacheManager
...
-                return Value;
+                var (Success, Value) = await TryGetAsync<T>(key);
+                if (Success)
+                {
+                    return Value;
+                }
+            }
+            catch(Exception ex) when (ex is not ArgumentNullException && _cacheOptions.DistributedCacheResilientPolicy == DistributedCacheResilientPolicy.FallbackToApi) 
+            {
+                _logger.LogInformation(ex, "Distributed cache is not available. Default DeliveryClient was used to get content from Delivery API.");
```
gormal, line 49: Is information the correct level?

Kontent.Ai.Delivery.Caching/DistributedCacheManager.cs
gormal: Oh, I see, that you made it optional in `AddDeliveryClientCache`, but is this ctor accessible from the customer project as well?

gormal (commented):
Looks great to me! I have some questions tho.

Kontent.Ai.Delivery.Caching/DistributedCacheManager.cs
dzmitryk-kontent-ai: I thought about it. It seems, that this ctor is available from the customer project. Nevertheless, in documentation we provide examples, which use:
- `CacheManagerFactory.Create()` method when customer registers multiple clients
- `AddDeliveryClientCache()` in ServiceCollectionExtensions, when customer registers a single client
Both these methods create entity of `DistributedCacheManager` class and have `loggerFactory` as optional parameter.
I don't know, whether some of our clients can use this ctor directly

dzmitryk-kontent-ai: Finally, I decided to make it optional in this ctor too. Just to be sure that I will not break someone's code

Kontent.Ai.Delivery.Caching/DistributedCacheManager.cs
gormal: Honestly, I don't know how to determine the correct severity of this issue, so let's go with warning and wait for some feedback :)

gormal (dismissed):
LGTM!

Simply007 (changes requested):
Just two parts

* Tests - I might shrink the `BrokenDistributedCache.cs` implementation to the test to a bare minimum (no check `if` statements and all async methods directly calling `Task.FromResult(NonAsyncMethod())`) - @gormal - do you agree?
* _This does not seem right after second look_ I might unify the new DeliveryCache options `DistributedCacheResilientPolicy ` with the `CacheType` 
    * Extend the description od `CacheTypeEnum.Distributed` with the explanation that is cache in unavailable - it will fail
    * Add something like `CacheTypeEnum.DistributedWIthApiFallback` (I am open to a better name) which would start fallback to the API if Cache is not available.

Simply007:
Do we even need to make this resilience configurable?

Simply007:
Another alternative to cut out ILoogerFactory logic from the PR and rely on user to register ILoogerFactory by himself and cut the option for resilience out and set default behavior to fallback with log.
* User knows about fail (because of logs) -> will be documented and do not get the exception.

gormal:
> Do we even need to make this resilience configurable?

When a customer's cache stops functioning, it is their decision whether to continue using DAPI (and incurring costs) or to crash the application and resolve the issue quickly.

> * Tests - I might shrink the `BrokenDistributedCache.cs` implementation to the test to a bare minimum (no check `if` statements and all async methods directly calling `Task.FromResult(NonAsyncMethod())`) - @gormal - do you agree?

Yeah, sounds great. 

> * Extend the description od `CacheTypeEnum.Distributed` with the explanation that is cache in unavailable - it will fail
> * Add something like `CacheTypeEnum.DistributedWIthApiFallback` (I am open to a better name) which would start fallback to the API if Cache is not available.

I would strongly advise against unifying these settings. It's easy to unify but very hard to separate in the future. Maybe there are some other options how to achieve the independence of these settings. 

> Another alternative to cut out ILoogerFactory logic from the PR and rely on user to register ILoogerFactory by himself and cut the option for resilience out and set default behavior to fallback with log.
> 
> * User knows about fail (because of logs) -> will be documented and do not get the exception.
> 
> Or add `WithLogger`/`WithLoggerFactory` similarly we have with `WithRetryPolicyProvider`, `WithPropertyManager` .

I like the `WithLoggerFactory` idea.

Simply007:
* let's keep `BrokenDistributedCache` as is
* keep the setting as it is for `CacheOptions` - lat' take a look at this on review, if something pop up 
* introduce `WithLoggerFactory` -> I am open to keeping it as a separate issue (but remove the current logic with login in this PR)

dzmitryk-kontent-ai:
@Simply007 , please go through the code, especially documentation changes. Since previous review I added `WithLoggerFactory` extension method which allows to register `ILoggerFactory` implementation during `DeliveryClient` creation

Simply007 (approved):
Great. Understandable from the documentation.

Good catch of the retry vs resilient in the docs - linking internal issue introduced it: KCL-1222

Issue #27, opened 2017-03-20
Generic filters


petrsvihlik:
Current situation: 
When I want to use e.g. `GreaterThanFilter` with `DateTime` I have to know how to format the `DateTime` so that the API accepts it. 

Proposed solution:
We should solve this in a generic way. The filters should accept `<T>` as a value and handle the formatting internally based on the type.

JanLenoch:
As an alternative to the solution proposed by @petrsvihlik , it could be implemented in the following simple way:

- The Filter class could have another constructor that takes `object value` instead of `string value`.
- In the code of that constructor, there could be an `if ... else` that converts the filter value to a proper string representation, based on the type information.

Like so:

    public Filter(string elementOrAttributePath, object value)
    {
        ElementOrAttributePath = elementOrAttributePath;

        if (value != null)
        {
            if (value is DateTime?)
            {
                DateTime dtValue = (DateTime)value;
                Value = dtValue.ToString("yyyy-MM-dd");
            }
            else if (value is IEnumerable<string>)
            {
                IEnumerable<string> ieValue = (IEnumerable<string>)value;
                Value = string.Join(",", ieValue);
            }
            // else if ...
            else
            {
                Value = value.ToString();
            }
        }
    }

Then, the signature of the existing constructors in covariant classes could be changed to accept `object value` (instead of `string value`), without losing backward compatibility.

Optional: We can enhance such code to also support our own Delivery SDK types (as mentioned in the table in https://forums.kenticocloud.com/discussion/31/strong-types-explained-code-generator/). Developers could then easily use e.g. the `MultipleChoiceOption` objects to pass as filtering values to the `ContainsFilter`.

petrsvihlik:
but we're open to any solutions...as long as they are backward compatible... :)

alesk-kentico:
I appreciate your effort, this issue has been mentioned several times and it should be resolved. However, we should not care about implementation yet, I consider API design and usage patterns more important. For Kentico developer the most important feature is to be able to easily discover what options are there. Let's admit it, IntelliSense is the primary way of finding out what the API can do.

The suggested approaches share the same weak spot – API does not tell you, which types of values can be used. Moreover, you can get wrong results when a value of unsupported type is converted to a string.

I propose a single solution – overloading methods. Just declare the relevant methods several times for each type that Delivery API supports. This way it will be very easy to see what types of parameters I can use and also it will be very difficult to make a mistake.

Enngage:
[...]

Issue #163, opened 2019-08-05
Add filters for `system` fields


petrsvihlik:
### Motivation

- Have a dedicated means of filtering by `system.type` (params string[]), `system.codename` and other system parameters, so that the developer does not have to type `system.type` in the [EqualsFilter](https://github.com/Kentico/delivery-sdk-net/blob/master/KenticoCloud.Delivery/QueryParameters/Filters/EqualsFilter.cs).

Parameters to cover according to the [documentation](https://developer.kenticocloud.com/reference#section-system-object-content-item-) **<parameter (constructor type)>**:

- id (Guid)
- codename (string)
- language (will be covered in #56)
- type (string)
- sitemap_locations (string[])
- last_modified (datetime)

Make use of strongly typed filters (#119).

For `system.type`, refactor the following occurrences:

- https://github.com/Kentico/delivery-sdk-net/blob/1167d6133847424a2ddf13473317557d4f1abf3c/KenticoCloud.Delivery/DeliveryClient.cs#L501
- https://github.com/Kentico/delivery-sdk-net/blob/1167d6133847424a2ddf13473317557d4f1abf3c/KenticoCloud.Delivery/DeliveryClient.cs#L512

pokornyd:
This is **partially addressed** in the modernized SDK (v19, `vnext` branch). The query API now has a fluent `IItemsFilterBuilder` with `System()` and `Element()` methods that handle the property path prefixing, along with typed filter operators (`string`, `double`, `DateTime`, `bool`) — this covers the generic filter functionality from PR #384:

```csharp
var result = await client.GetItems<Article>()
    .Where(f => f
        .System("codename").IsIn("on-roasts", "which-brewing-fits-you")
        .System("last_modified").IsGreaterThan(DateTime.UtcNow.AddDays(-30))
        .System("collection").IsEqualTo("blog"))
    .ExecuteAsync();
```

Additionally, typed queries like `GetItems<Article>()` **automatically** add the `system.type` filter when an `ITypeProvider` is registered — so the most common system filter doesn't need to be specified at all.

### What's still missing

The original issue asked for **dedicated system field helpers** — I assume something along the lines of `.WithType("article")`, `.WithCollection("blog")` — so developers don't have to know the property name strings. I'll consider adding this.

Full documentation: [Advanced Filtering Guide](https://github.com/kontent-ai/delivery-sdk-net/blob/vnext/docs/advanced-filtering.md)

---

[...]

Issue #56, opened 2017-06-28
Add filters for system attributes (language, ...)


petrsvihlik:

kordys:
Hey @petrsvihlik! The language filter should add, based on [documentation](https://developer.kenticocloud.com/v1/docs/understanding-language-fallbacks), an additional query string parameter to disable the fallback. Which means the GetQueryStringParameter method can return in fact two parameters. The question is if we want to change the return type of GetQueryStringParameter method to be an array of query string parameters or we are ok with returning multiple parameters as a single string... I can imagine there will be more parameters with multiple query string parameters in the future.

petrsvihlik:
hey @kordys . I'm quite ok with both. However, I vote for returning an array/collection of key-value pairs as there are some other advantages. For instance, you wouldn't have to call `Uri.EscapeDataString` in every single `IQueryParameter.GetQueryStringParameter()` implementation. Moreover, I think there is some built-in support for generating querystrings from a collection in .NET Core.

pokornyd:
This is **fully addressed** in the modernized SDK (v19, `vnext` branch). The old `LanguageParameter` class with its constructor-based approach has been replaced by a fluent `WithLanguage` method that has built-in support for ignoring language fallbacks via a `LanguageFallbackMode` enum:

```csharp
// With fallbacks (default behavior)
var result = await client.GetItems<Article>()
    .WithLanguage("en-US")
    .ExecuteAsync();

// Without fallbacks — only returns items that exist in the requested language
var result = await client.GetItems<Article>()
    .WithLanguage("en-US", LanguageFallbackMode.Disabled)
    .ExecuteAsync();
```

When `LanguageFallbackMode.Disabled` is passed, the SDK automatically adds the `system.language` filter alongside the `language` parameter — exactly what the original issue and the [API documentation](https://kontent.ai/learn/docs/apis/openapi/delivery-api/#section/Localization/Ignoring-language-fallbacks) describe.

No need for separate constructors or manually combining parameters — it's a single method call with an optional enum.

---

The SDK is currently pre-released as 19.0.0-rc3, with production release planned for early April, at which point I'll close this issue. It's compatible with the latest beta of [model generator](https://github.com/kontent-ai/model-generator-net). Sync API features are newly released as a [standalone SDK](https://github.com/kontent-ai/sync-sdk-net).

PR #384, opened 2023-10-17
Generic Filters

### LINKED ISSUE #119: Spike:Make Filters and Parameters less error-prone

nkooman:
### Motivation

Fixes #119

I did not add the specific `DateTime` and `decimal` constructors. I have also kept the original method signatures for all of the filters for backward compatibility.
The issue (#119) calls out that all filters should implement `DateTime` and `decimal` constructors, then utilize a new constructor in `Filter` to accept `object`s as values. The original issue (#27), has a proposed solution in such a way; however, doing so defeats the purpose of creating a generic filter in the first place. Since, if we were to go the `object` route, all values would get converted to a `string`, removing the need for a generic and for `GetQueryStringParameter` to handle types in a `switch`.
Instead, I opted to go the route of the generic filter and implemented a switch statement to interpret `string`, `DateTime`, and numeric types (`int`, `decimal`, `float`, etc.). The downside here is that we lose the specific constructors for `DateTime` and `decimal` values.

### Checklist

- [x] Tests are passing

### How to test

Create queries using the new filters and confirm that results are correctly filtered.


nkooman:
I've been looking into some of the other issues and think it would be better if we combined #56, #163, and #384 into a single issue. Then we can refactor the filters so that they support a Fluent API structure.
@pokornyd Looking for some thoughts on this idea before I close this pull request.

pokornyd:
Hello @nkooman and thank you for your contribution, forgive the rather late response. I'll review your proposed changes, as well as the issues you suggested for merging with this PR and get back to you with any comments, thanks for your patience.

PR #403, opened 2025-02-28
Add support for used in endpoints

winklertomas:
### Motivation

Add support for new used-in delivery API feature. 

### Checklist

- [ ] Tests are passing

### Release date

Please do not release before the used-in is released in the production. The early access should be released in the middle of March, we will let you know.

### DIFF
--- a/Kontent.Ai.Delivery.Rx/DeliveryObservableProxy.cs
+++ b/Kontent.Ai.Delivery.Rx/DeliveryObservableProxy.cs
@@ -2,6 +2,8 @@
 using System.Collections.Generic;
 using System.Reactive.Disposables;
 using System.Reactive.Linq;
+using System.Reactive;
+using System.Threading.Tasks;
 using Kontent.Ai.Delivery.Abstractions;
 
 namespace Kontent.Ai.Delivery.Rx
@@ -104,8 +106,7 @@ namespace Kontent.Ai.Delivery.Rx
         /// <returns>The <see cref="IObservable{T}"/> that represents the content items. If no query parameters are specified, all content items are returned.</returns>
         public IObservable<T> GetItemsFeedObservable<T>(IEnumerable<IQueryParameter> parameters) where T : class
         {
-            var feed = DeliveryClient?.GetItemsFeed<T>(parameters);
-            return feed == null ? null : EnumerateFeed(feed)?.ToObservable();
+            return DeliveryClient?.GetItemsFeed<T>(parameters).ToObservable();
         }
 
         /// <summary>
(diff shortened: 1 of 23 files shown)


Kontent.Ai.Delivery/DeliveryClient.cs
```diff
@@ -363,6 +364,48 @@ public async Task<IDeliverySyncResponse> GetSyncAsync(string continuationToken)
             return new DeliverySyncResponse(response, itemModels);
         }
 
+        /// <summary>
+        /// Returns a feed that is used to traverse through strongly typed parent content items matching the optional filtering parameters.
+        /// </summary>
+        /// <param name="codename">The codename of a content item.</param>
+        /// <param name="parameters">A collection of query parameters for filtering.</param>
+        /// <returns>The <see cref="DeliveryUsedInItems"/> instance that can be used to enumerate through content item parents for the specified item codename. If no query parameters are specified, default language parents are enumerated.</returns>
+        public IDeliveryItemsFeed<IUsedInItem> GetItemUsedIn(string codename, IEnumerable<IQueryParameter> parameters = null)
```
pokornyd, line 373: existing methods accepting a `codename` parameter have null handling in place. consider adding it to all usedIn methods as well:

```csharp
if (codename == null)
{
    throw new ArgumentNullException(nameof(codename), "The codename of a content item is not specified.");
}
```

Kontent.Ai.Delivery/DeliveryClient.cs
```diff
@@ -363,6 +364,48 @@ public async Task<IDeliverySyncResponse> GetSyncAsync(string continuationToken)
...
+
+        /// <summary>
+        /// Returns a feed that is used to traverse through strongly typed parent content items matching the optional filtering parameters.
+        /// </summary>
+        /// <param name="codename">The codename of an asset.</param>
+        /// <param name="parameters">A collection of query parameters for filtering.</param>
+        /// <returns>The <see cref="DeliveryUsedInItems"/> instance that can be used to enumerate through asset parents for the specified asset codename. If no query parameters are specified, default language parents are enumerated.</returns>
+        public IDeliveryItemsFeed<IUsedInItem> GetAssetUsedIn(string codename, IEnumerable<IQueryParameter> parameters = null)
+        {
+            ValidateUsedInParameters(parameters);
```
pokornyd, line 389: also null handling:

```csharp
if (codename == null)
{
    throw new ArgumentNullException(nameof(codename), "The codename of an asset is not specified.");
}
```

Kontent.Ai.Delivery.Rx/DeliveryObservableProxy.cs
```diff
@@ -211,6 +246,16 @@ public IObservable<ILanguage> GetLanguagesObservable(params IQueryParameter[] pa
 
         #endregion
         #region "Private methods"
+        private static IEnumerable<T> EnumerateFeed<T>(IDeliveryItemsFeed<T> feed) where T : class
+        {
```
pokornyd, line 250: this method is synchronous but invokes an async method and waits for its result, effectively blocking threads. not sure if this is a valid scenario under the circumstances, but I'm not familiar with Rx so I asked AI about it. 

the fix is simple, make the method async and have it return `IAsyncEnumerable`:

```csharp
private static async IAsyncEnumerable<T> EnumerateFeed<T>(IDeliveryItemsFeed<T> feed) where T : class
{
    while (feed.HasMoreResults)
    {
        var batch = await feed.FetchNextBatchAsync();
        foreach (var contentItem in batch.Items)
        {
            yield return contentItem;
        }
    }
}
```

however, related methods would need to be updated as well.

again, I'm not sure if thread locking or even deadlocks can be a thing in our scenario. also, Rx is used much less, compared to async, so I leave this at your decision.

pokornyd (changes requested):
add null handling and consider changing `EnumerateFeed` to `async`

pokornyd (approved):

